PR #11, opened 2021-04-19
Merge history from StatsFuns

devmotion:
This PR adds the git history from StatsFuns extracted with
```shell
git filter-repo --path src/constants.jl --path src/basicfuns.jl --path test/basicfuns.jl
```
and merges it with the modified docstrings in this repo. Additionally, the list of exports is synced with StatsFuns and now additionally the constants and aliases are exported.


nalimilan:
Thanks. Where should one look to see the diff between StatsFuns and LogExpFunctions? I thought it would be in the merge commit, but it seems to be empty?

devmotion:
It seems the merge commit was created automatically when I pulled in LogExpFunctions... The last commits show how I merged the different changes manually: https://github.com/JuliaStats/LogExpFunctions.jl/pull/11/files/e1fe1a4a979ce2e3fe48349069bd715383718e62..603a364503b5df08cbb3eaf574d6be536535d478 However, I assume it is easier to keep track of the changes by inspecting them one after another.

nalimilan:
That's weird. In case of conflicts I would have expected git to ask you to resolve them manually. How about squashing the relevant commits into the merge commit? That should be clearer when looking at the git history in the future.

devmotion:
I could squash the last commits but I am not sure if it is helpful to squash them into the merge commit. The merge commit really just introduced the history and the files from LogExpFunctions but there were no actual merge conflicts to resolve. In this step I just ended up with duplicate implementations in different files (e.g. `src/LogExpFunctions.jl` and `src/basicfuns.jl`) which I then cleaned in the following commits. So it might be even more confusing if all these fixes are applied in the merge commit?

nalimilan:
Ah got it, so that's because the files have different names. Maybe at least make it clear in commit messages which ones are just moving code around without modifying anything? Otherwise it's super hard to spot actual changes.

nalimilan (approved):
Let's fix this at last! :-)

PR #16, opened 2021-04-20
Update softmax with parameterization

bdeonovic:
parameterize softmax to be usable with AD.

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -260,4 +260,4 @@ $(SIGNATURES)
 Return the [`softmax transformation`](https://en.wikipedia.org/wiki/Softmax_function)
 applied to `x`.
 """
-softmax(x::AbstractArray{<:Real}) = softmax!(similar(x, float(eltype(x))), x)
+softmax(x::AbstractArray{<:Real}) = softmax!(similar(x, eltype(x)), x)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -240,8 +240,9 @@ function softmax!(r::AbstractArray{<:Real}, x::AbstractArray{<:Real})
     @inbounds for i = 1:n
         s += (r[i] = exp(x[i] - u))
     end
+    invs = inv(s)
     @inbounds for i = 1:n
-        r[i] *= inv(s)
+        r[i] *= invs
     end
     return r
 end
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -241,7 +241,7 @@ function softmax!(r::AbstractArray{<:Real}, x::AbstractArray{<:Real})
         s += (r[i] = exp(x[i] - u))
     end
     @inbounds for i = 1:n
-        r[i] /= s
+        r[i] *= inv(s)
     end
     return r
 end
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -261,4 +261,4 @@ $(SIGNATURES)
 Return the [`softmax transformation`](https://en.wikipedia.org/wiki/Softmax_function)
 applied to `x`.
 """
-softmax(x::AbstractArray{<:Real}) = softmax!(similar(x, Float64), x)
+softmax(x::AbstractArray{R<:Real}) = softmax!(similar(x, R), x)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -232,7 +232,7 @@ That is, `r` is overwritten with `exp.(x)`, normalized to sum to 1.
 
 See the [Wikipedia entry](https://en.wikipedia.org/wiki/Softmax_function)
 """
-function softmax!(r::AbstractArray{R}, x::AbstractArray{T}) where {R<:Real, T<:Real}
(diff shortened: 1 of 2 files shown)


src/basicfuns.jl
```diff
@@ -261,4 +261,4 @@ $(SIGNATURES)
 Return the [`softmax transformation`](https://en.wikipedia.org/wiki/Softmax_function)
 applied to `x`.
 """
-softmax(x::AbstractArray{<:Real}) = softmax!(similar(x, Float64), x)
+softmax(x::AbstractArray{R<:Real}) = softmax!(similar(x, R), x)
```
devmotion, line 264: I suggest
```suggestion
softmax(x::AbstractArray{<:Real}) = softmax!(similar(x, float(eltype(x))), x)
```
This avoids the type parameter (and the missing `where` clause) and should work with integervalued arrays as well.

devmotion (changes requested):

src/basicfuns.jl
devmotion: Can you also add a test to `test/basicfuns.jl` that checks that it works for integer-valued arrays? And maybe Float32 as expected, i.e., returns an array of element type Float32?

devmotion:
Depending on your use case, you might need to improve the implementation of `softmax!` as well (e.g., I guess `<:AbstractFloat` is quite restrictive and `s = 0.` should probably not be hardcoded as Float64).

bdeonovic:
Do you have any comments on why softmax! was written the way it was? Is there a point to using `inv` rather than just dividing by the sum `s`?

devmotion:
> Do you have any comments on why softmax! was written the way it was? Is there a point to using `inv` rather than just dividing by the sum `s`?

No, I don't know why it was written in this way. I assume that it uses `inv` for performance reasons: usually multiplication is faster than division.

src/basicfuns.jl
```diff
@@ -232,19 +232,18 @@ That is, `r` is overwritten with `exp.(x)`, normalized to sum to 1.
 
 See the [Wikipedia entry](https://en.wikipedia.org/wiki/Softmax_function)
 """
-function softmax!(r::AbstractArray{R}, x::AbstractArray{T}) where {R<:AbstractFloat,T<:Real}
+function softmax!(r::AbstractArray{R}, x::AbstractArray{T}) where {R<:Real, T<:Real}
```
devmotion, line 235: Can we use
```suggestion
function softmax!(r::AbstractArray{<:Real}, x::AbstractArray{<:Real})
```
here?

src/basicfuns.jl
```diff
@@ -232,19 +232,18 @@ That is, `r` is overwritten with `exp.(x)`, normalized to sum to 1.
...
     n = length(x)
     length(r) == n || throw(DimensionMismatch("Inconsistent array lengths."))
     u = maximum(x)
-    s = 0.
+    s = zero(R)
```
devmotion, line 239:
```suggestion
    s = zero(eltype(r))
```

src/basicfuns.jl
```diff
@@ -232,19 +232,18 @@ That is, `r` is overwritten with `exp.(x)`, normalized to sum to 1.
...
     @inbounds for i = 1:n
         s += (r[i] = exp(x[i] - u))
     end
-    invs = convert(R, inv(s))
```
devmotion, line 243: I guess it is faster to compute the inverse only once but it would be interesting to check if this is actually true. In any case, I don't think it is needed to convert the inverse since `s` is initialized based on `eltype(r)` now.

test/basicfuns.jl
```diff
@@ -149,7 +149,15 @@ end
...
     softmax!(x)
     @test x ≈ r
+    
+    x = [1, 2, 3]
+    r = exp.(x) ./ sum(exp.(x))
+    @test softmax(x) ≈ r
+    
+    x = [1//2, 2//3, 3//4]
+    r = exp.(x) ./ sum(exp.(x))
+    @test softmax(x) ≈ r
```
devmotion, line 162: Can you test the element type of `softmax(x)` in these examples and add a test with a `Vector{Float32}`?

src/basicfuns.jl
bdeonovic: Is this just more succinct?

test/basicfuns.jl
bdeonovic: for rational and Int64 softmax will output Float64 (because of the float(eltype(x)) call) for Float32 is this acceptable behavior?

devmotion: Yep, this is the expected behaviour. The Float32 test is the most interesting one here I assume since currently it would just output a vector of Float64 in this case.

devmotion (commented):
OK, looks good! If you revert back to using `invs = inv(s)`, then it should be good to go and I'll merge and make a new release.

bdeonovic:
thanks for your patience

devmotion (approved):
Thanks a lot for the PR!

devmotion:
A new release with the fix should be available soon: https://github.com/JuliaRegistries/General/pull/34861

Issue #12, opened 2021-01-23
Softmax segfaults with offset arrays


yurivish:
Due to improper inbounds annotations, `softmax` segfaults when called on an `OffsetArray`:

```julia
yurivish@Compy ~ % julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.3 (2020-11-09)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using StatsFuns, OffsetArrays

julia> softmax(OffsetArray(collect(1:11), -5:5))
zsh: segmentation fault  julia
yurivish@Compy ~ %
```

https://github.com/JuliaStats/StatsFuns.jl/blob/8dfda2c0ee33d5f85eca5c039d31d85c90f363f2/src/basicfuns.jl#L330-L342

```
julia> versioninfo()
Julia Version 1.5.3
Commit 788b2c77c1 (2020-11-09 13:37 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
```

PR #17, opened 2021-04-21
Allow general indices in `softmax!`

devmotion:
This PR fixes https://github.com/JuliaStats/LogExpFunctions.jl/issues/12. Additionally, it allows to use `softmax` and `softmax!` with `CuArray`s (tested locally but not included in the testset).

### DIFF
--- a/src/LogExpFunctions.jl
+++ b/src/LogExpFunctions.jl
@@ -2,7 +2,7 @@ module LogExpFunctions
 
 using DocStringExtensions: SIGNATURES
 using Base: Math.@horner, @irrational
-using LinearAlgebra: LinearAlgebra
+import LinearAlgebra
 
 export loghalf, logtwo, logπ, log2π, log4π
 export xlogx, xlogy, logistic, logit, log1psq, log1pexp, log1mexp, log2mexp, logexpm1,
--- a/src/LogExpFunctions.jl
+++ b/src/LogExpFunctions.jl
@@ -2,6 +2,7 @@ module LogExpFunctions
 
 using DocStringExtensions: SIGNATURES
 using Base: Math.@horner, @irrational
+using LinearAlgebra: LinearAlgebra
 
 export loghalf, logtwo, logπ, log2π, log4π
 export xlogx, xlogy, logistic, logit, log1psq, log1pexp, log1mexp, log2mexp, logexpm1,
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -233,17 +233,12 @@ That is, `r` is overwritten with `exp.(x)`, normalized to sum to 1.
 See the [Wikipedia entry](https://en.wikipedia.org/wiki/Softmax_function)
 """
 function softmax!(r::AbstractArray{<:Real}, x::AbstractArray{<:Real})
-    n = length(x)
-    length(r) == n || throw(DimensionMismatch("Inconsistent array lengths."))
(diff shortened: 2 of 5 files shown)


src/LogExpFunctions.jl
```diff
@@ -2,6 +2,7 @@ module LogExpFunctions
 
 using DocStringExtensions: SIGNATURES
 using Base: Math.@horner, @irrational
+using LinearAlgebra: LinearAlgebra
```
nalimilan, line 5: This is weird. Is this what you want?
```suggestion
import LinearAlgebra
```

src/basicfuns.jl
```diff
@@ -233,17 +233,12 @@ That is, `r` is overwritten with `exp.(x)`, normalized to sum to 1.
...
-    @inbounds for i = 1:n
-        s += (r[i] = exp(x[i] - u))
-    end
-    invs = inv(s)
-    @inbounds for i = 1:n
-        r[i] *= invs
+    map!(r, x) do xi
+        return exp(xi - u)
     end
+    LinearAlgebra.lmul!(inv(sum(r)), r)
```
nalimilan, line 241: Isn't it slower to compute the sum again at the end? We have again the same issue that `CuArray`'s requirements are not necessarily the best for `Array`.

test/basicfuns.jl
```diff
@@ -149,22 +149,36 @@ end
...
+    @test eltype(s) === Float64
     
-    x = Float32[1, 2, 3]
-    r = exp.(x) ./ sum(exp.(x))
-    @test softmax(x) ≈ r
-    @test eltype(softmax(x)) == Float32
+    # non-standard indices: #12
+    x = OffsetArray(1:3, -2:0)
+    s = softmax(x)
+    @test s isa OffsetArray
```
nalimilan, line 182: Also check indices?

src/LogExpFunctions.jl
devmotion: It's how BlueStyle and JuliaFormatter require you to write it (`using X: X` instead of `import X`). I am just involved in too many projects that use BlueStyle currently and therefore automatically did it here as well, even though there's no style enforced. Do you want me to change it?

src/basicfuns.jl
devmotion: I benchmarked many different implementations such as the current one, iteration with `eachindex`, `map!`, broadcasting (with `vec` to avoid dimension mismatches), with additional `sum` and without, and also more obscure ones such as `map!` closed over a summation variable. Both for arrays with 10 and 1000 elements the timings were quite similar and in most cases I didn't notice any clear performance regressions. Surprisingly, in some cases actually the explicit `sum` was faster - maybe it is easier e.g. for the compiler to optimize the for loop without manual summation?

So my impression was that the `sum` does not cause any clear performance regressions and therefore I went for this more general approach. The finding also didn't seem very surprising since intuitively exponentiation is the most expensive part of the softmax function.

src/LogExpFunctions.jl
devmotion: Sure, I don't mind.

test/basicfuns.jl
devmotion: I added a check of the indices and the type of the elements.

src/basicfuns.jl
nalimilan: Cool, I guess the simple version is OK then.

nalimilan (approved):

PR #19, opened 2021-05-18
Generalize `logsumexp` to complex numbers

devmotion:


test/basicfuns.jl
```diff
@@ -140,10 +154,23 @@ end
     @test isnan(logsumexp([NaN, 9.0]))
     @test isnan(logsumexp([NaN, Inf]))
     @test isnan(logsumexp([NaN, -Inf]))
+    @test isnan(logsumexp(Complex{Float64}[NaN, 9.0]))
+    @test isnan(logsumexp(Complex{Float64}[NaN, Inf]))
+    @test isnan(logsumexp(Complex{Float64}[NaN, -Inf]))
```
nalimilan, line 159: Also test `NaN` in the imaginary part?

src/logsumexp.jl
```diff
@@ -61,36 +61,59 @@ _logsumexp_onepass_reduce(X, ::Base.EltypeUnknown) = reduce(_logsumexp_onepass_o
...
 end
+_logsumexp_onepass_op(x1::Number, x2::Number) = _logsumexp_onepass_op(promote(x1, x2)...)
 
 # reduce a number and a partial sum
-function _logsumexp_onepass_op(x, (xmax, r)::Tuple)
-    a = x == xmax ? zero(x - xmax) : -abs(x - xmax)
-    if x > xmax
-        _xmax = oftype(a, x)
-        _r = (r + one(r)) * exp(a)
+function _logsumexp_onepass_op(x::Number, (xmax, r)::Tuple{<:Number,<:Number})
```
nalimilan, line 79: Use the short function form (here and below)?

src/logsumexp.jl
```diff
@@ -61,36 +61,59 @@ _logsumexp_onepass_reduce(X, ::Base.EltypeUnknown) = reduce(_logsumexp_onepass_o
...
 
 # reduce two numbers
-function _logsumexp_onepass_op(x1, x2)
-    a = x1 == x2 ? zero(x1 - x2) : -abs(x1 - x2)
-    xmax = x1 > x2 ? oftype(a, x1) : oftype(a, x2)
+function _logsumexp_onepass_op(x1::T, x2::T) where {T<:Number}
+    xmax, a = if x1 == x2
+        # handle `Inf` etc. correctly
+        x2, zero(x1 - x2)
+    elseif real(x1) > real(x2)
```
nalimilan, line 68: I haven't thought too deeply about this, but how does it work to check only the real part? :-)

src/logsumexp.jl
```diff
@@ -61,36 +61,59 @@ _logsumexp_onepass_reduce(X, ::Base.EltypeUnknown) = reduce(_logsumexp_onepass_o
 ## Reductions for one-pass algorithm: avoid expensive multiplications if numbers are reduced
 
 # reduce two numbers
-function _logsumexp_onepass_op(x1, x2)
-    a = x1 == x2 ? zero(x1 - x2) : -abs(x1 - x2)
-    xmax = x1 > x2 ? oftype(a, x1) : oftype(a, x2)
```
nalimilan, line 66: So the `oftype` isn't needed now?

test/basicfuns.jl
devmotion: OK, I'll add it to the tests :+1:

src/logsumexp.jl
devmotion: You mean just `_logsumexp_onepass_op(x::Number, (xmax, r)::Tuple{...}) = ...`?

nalimilan: Yes.

src/logsumexp.jl
devmotion: The real part determines the absolute value of `exp(z)`, and hence it is used to determine which term is used to scale the sum of exponentials. We do not scale with the real part though but the full complex number since this allows us to use `log1p` for computing the logarithm of the scaled exponentials in the final reducing step, which avoids underflow.

devmotion: This is also explained in https://scicomp.stackexchange.com/questions/34273/log-sum-exp-trick-for-signed-complex-numbers#comment69041_34275.

src/logsumexp.jl
devmotion: No, it is not needed anymore since the terms in both branches are guaranteed to be of the same type since the relevant arguments are ensured to be of the same type. I thought in this way the code would be easier to write and to read.

nalimilan: But `oftype(x1, NaN)` may not be of the same type as `x1 - x2`, right? This should be `oftype(x1-x2, NaN)`.

devmotion: `x1` and `x2` are of the same type, so `oftype(x1, NaN)` should be of the same type as `x1 - x2`.

nalimilan (approved):

PR #21, opened 2021-06-16
Fix `logsubexp(-Inf, -Inf)`

devmotion:


devmotion:
Bump :slightly_smiling_face:

test/basicfuns.jl
```diff
@@ -135,7 +135,7 @@ end
     end
 
     @test isnan(logsubexp(Inf, Inf))
-    @test isnan(logsubexp(-Inf, -Inf))
+    @test logsubexp(-Inf, -Inf) ≡ -Inf
```
nalimilan, line 138: Maybe test all combinations of `Inf` and `-Inf`? Also, why isn't `NaN` as second argument tested?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "LogExpFunctions"
 uuid = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 authors = ["StatsFun.jl contributors, Tamas K. Papp <tkpapp@gmail.com>"]
-version = "0.2.4"
+version = "0.2.5"
```
nalimilan, line 4: Technically I guess this PR should be considered as breaking, right? Though in practice I agree it would be annoying to tag a breaking release so maybe that's OK. (If we had reached 1.0 we could tag a minor release...)

devmotion: I think this is just a bugfix (the current behaviour is incorrect) and hence non-breaking.

test/basicfuns.jl
devmotion: I added more tests.

nalimilan (approved):

Project.toml
tpapp: I agree, this is a bugfix.

PR #22, opened 2021-07-04
Use IrrationalConstants and do not export logarithmic constants

devmotion:
The constants in StatsFuns were moved to [IrrationalConstants](https://github.com/JuliaMath/IrrationalConstants.jl) such that other packages such as e.g. SpecialFunctions or DiffRules can make use of them without having to depend on StatsFuns. This PR ~~reexports the logarithmic constants in IrrationalConstants and~~ updates the documentation (also adds a missing docstring for `logsubexp` and fixes doctests which could be moved to a separate PR).

**Edit:** This PR removes the definition of the logarithmic constants in LogExpFunctions and uses the definitions in IrrationalConstants instead but does not reexport the constants defined in IrrationalConstants. Users and downstream packages should depend on IrrationalConstants directly.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,10 +5,12 @@ version = "0.2.5"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+IrrationalConstants = "92d709cd-6900-40b7-9082-c6be49f344b6"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 
 [compat]
 DocStringExtensions = "0.8"
+IrrationalConstants = "0.1"
 julia = "1"
 
 [extras]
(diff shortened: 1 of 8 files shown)


tpapp:
I think that moving those constants to their own package is a great idea, but I am not sure we should reexport them. This package is merely a *user* of those constants, those who need them should just be `using IrrationalConstants`, too.

devmotion:
I don't mind removing the constants from LogExpFunctions (I mean it's also called "Functions" :smile:), I just assumed a non-breaking change would be easier. We could also make a non-breaking release but deprecate the constants with `@deprecate_binding` and then remove them in the next breaking release.

tpapp:
I would be OK with just making a breaking change, but deprecation and then breaking is fine too.

devmotion:
OK, I removed the constants and prepared a breaking release.

tpapp:
I think we still need to be `using IrrationalConstants` though, as they are used in the code.

devmotion:
Ah true, I missed that it's not only exported :smile:

devmotion:
It's fixed now.

tpapp (approved):
LGTM. Before merging, please change the title and the PR explanation to reflect the end result, particularly because the release notes created by TagBot should be informative about the change and what the user should do (import IrrationalConstants directly if needed).

devmotion:
Done :+1:

PR #25, opened 2021-09-05
Add ChainRules definitions

devmotion:
This PR adds ChainRules definitions for the derivatives of the functions in LogExpFunctions.

Currently, Zygote contains some reverse-mode differentiation rules for LogExpFunctions (https://github.com/FluxML/Zygote.jl/blob/master/src/lib/logexpfunctions.jl) but it does not cover all functions in the package and is Zygote-specific, i.e., it is not used by any other AD system. ChainRules allows us to define both forward- and reverse-mode derivatives (for all functions) in an AD system independent way. I think both the different AD backends and LogExpFunctions would benefit if a complete set of ChainRules definitions for LogExpFunctions would be added in a central Zygote-independent place.

Since ChainRulesCore 1 is a very lightweight package (see [@oxinabox's comment](https://github.com/JuliaStats/StatsFuns.jl/pull/106#issuecomment-905063473) for more details) and Pkg does not support conditional dependencies yet, I would argue that currently the best approach is to add these definitions to LogExpFunctions. ChainRulesCore is already widely used in the package ecosystem (1566 dependents as of today: https://juliahub.com/ui/Packages/ChainRulesCore/G6ax7/1.3.1?t=2), e.g., it is a dependency of SpecialFunctions and recently became a dependency of StatsFuns and Distributions.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,21 +1,25 @@
 name = "LogExpFunctions"
 uuid = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 authors = ["StatsFun.jl contributors, Tamas K. Papp <tkpapp@gmail.com>"]
-version = "0.3.0"
+version = "0.3.1"
 
 [deps]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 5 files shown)


tpapp (approved):
LGTM.

PR #26, opened 2021-09-05
Use ternary operator instead of ifelse

devmotion:
In benchmarks of https://github.com/JuliaStats/StatsBase.jl/pull/714 and the examples shown below it seems that on my computer the ternary operator is sometimes faster than `ifelse` (which evaluates all branches) and never slower:

With the latest release:
```julia
julia> using LogExpFunctions, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> out = Vector{Float64}(undef,100_000);

julia> x, y = rand(100_000), rand(100_000);

julia> @btime map!(xlogx, $out, $x);
  717.358 μs (0 allocations: 0 bytes)

julia> @btime map!(xlogy, $out, $x, $y);
  780.322 μs (0 allocations: 0 bytes)

julia> @btime map!(logistic, $out, $x);
  745.426 μs (0 allocations: 0 bytes)

julia> @btime map!(logaddexp, $out, $x, $y);
  2.502 ms (0 allocations: 0 bytes)

julia> @btime sum(xlogx, $x); # similar to https://github.com/JuliaStats/StatsBase.jl/pull/714
  780.203 μs (0 allocations: 0 bytes)
```

With this PR:
```julia
julia> using LogExpFunctions, BenchmarkTools, Random

julia> Random.seed!(1234);

julia> out = Vector{Float64}(undef,100_000);

julia> x, y = rand(100_000), rand(100_000);

julia> @btime map!(xlogx, $out, $x);
  703.294 μs (0 allocations: 0 bytes)

julia> @btime map!(xlogy, $out, $x, $y);
  780.819 μs (0 allocations: 0 bytes)

julia> @btime map!(logistic, $out, $x);
  717.576 μs (0 allocations: 0 bytes)

julia> @btime map!(logaddexp, $out, $x, $y);
  2.507 ms (0 allocations: 0 bytes)

julia> @btime sum(xlogx, $x);
  762.913 μs (0 allocations: 0 bytes)
```

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -11,7 +11,7 @@ julia> xlogx(0)
 """
 function xlogx(x::Number)
     result = x * log(x)
-    ifelse(iszero(x), zero(result), result)
+    return iszero(x) ? zero(result) : result
 end
 
 """
@@ -26,7 +26,7 @@ julia> xlogy(0, 0)
 """
 function xlogy(x::Number, y::Number)
     result = x * log(y)
-    ifelse(iszero(x) && !isnan(y), zero(result), result)
+    return iszero(x) && !isnan(y) ? zero(result) : result
 end
 
 # The following bounds are precomputed versions of the following abstract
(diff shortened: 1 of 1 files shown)


nalimilan (approved):
Thanks! AFAIK with `?` LLVM will decide whether to evaluate all branches or not, while with `ifelse` it will always evaluate. So if we trust LLVM `?` is generally better.

PR #28, opened 2021-09-09
Add `dims` keyword argument to `softmax` and `softmax!`

devmotion:
This PR adds a `dims` keyword argument to `softmax` and `softmax!`. The current behaviour (normalizing the whole array and neglecting mismatching dimensions) is preserved with the default value `dims=:`.

The PR also removes the ChainRules definition for `softmax!` and instead adds one for `softmax` (with `dims` keyword argument). Apparently, currently no rules should be defined for mutating functions (https://juliadiff.org/ChainRulesCore.jl/stable/writing_good_rules.html#Which-functions-need-rules?), so this is a bugfix.

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -1,11 +1,42 @@
 # This file is machine-generated - editing it directly is not advised
 
+[[ANSIColoredPrinters]]
+git-tree-sha1 = "574baf8110975760d391c710b6341da1afa48d8c"
+uuid = "a4c015fc-c6ff-483c-b24f-f7ea428134e9"
+version = "0.0.1"
+
+[[ArgTools]]
+uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+
+[[Artifacts]]
+uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
+
 [[Base64]]
 uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+
+[[ChainRulesCore]]
+deps = ["Compat", "LinearAlgebra", "SparseArrays"]
+git-tree-sha1 = "30ee06de5ff870b45c78f529a6b093b3323256a3"
+uuid = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+version = "1.3.1"
+
+[[Compat]]
+deps = ["Base64", "Dates", "DelimitedFiles", "Distributed", "InteractiveUtils", "LibGit2", "Libdl", "LinearAlgebra", "Markdown", "Mmap", "Pkg", "Printf", "REPL", "Random", "SHA", "Serialization", "SharedArrays", "Sockets", "SparseArrays", "Statistics", "Test", "UUIDs", "Unicode"]
+git-tree-sha1 = "4866e381721b30fac8dda4c8cb1d9db45c8d2994"
+uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
+version = "3.37.0"
 
 [[Dates]]
 deps = ["Printf"]
 uuid = "ade2ca70-3891-5945-98fb-dc099432e06a"
+
+[[DelimitedFiles]]
+deps = ["Mmap"]
+uuid = "8bb1440f-4735-579b-a4ab-409b98df4dab"
+
+[[Distributed]]
+deps = ["Random", "Serialization", "Sockets"]
+uuid = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 
 [[DocStringExtensions]]
 deps = ["LibGit2"]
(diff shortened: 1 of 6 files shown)


src/basicfuns.jl
```diff
@@ -239,18 +239,13 @@ $(SIGNATURES)
...
-    length(r) == length(x) || throw(DimensionMismatch("inconsistent array lengths"))
-    u = maximum(x)
-    map!(r, x) do xi
-        return exp(xi - u)
-    end
-    LinearAlgebra.lmul!(inv(sum(r)), r)
-    return r
+function softmax!(r::AbstractArray{<:Real}, x::AbstractArray{<:Real}; dims=:)
+    return _softmax!(r, x, dims)
 end
```
nalimilan, line 249:
```suggestion
softmax!(r::AbstractArray{<:Real}, x::AbstractArray{<:Real}; dims=:) =
    _softmax!(r, x, dims)
```

src/basicfuns.jl
```diff
@@ -259,12 +254,38 @@ $(SIGNATURES)
...
 """
 $(SIGNATURES)
 
 Return the [`softmax transformation`](https://en.wikipedia.org/wiki/Softmax_function)
 applied to `x`.
 """
-softmax(x::AbstractArray{<:Real}) = softmax!(similar(x, float(eltype(x))), x)
+function softmax(x::AbstractArray{<:Real}; dims=:)
+    return softmax!(similar(x, float(eltype(x))), x; dims=dims)
+end
```
nalimilan, line 267:
```suggestion
softmax(x::AbstractArray{<:Real}; dims=:) =
    softmax!(similar(x, float(eltype(x))), x; dims=dims)
```

src/basicfuns.jl
```diff
@@ -259,12 +254,38 @@ $(SIGNATURES)
 Return the [`softmax transformation`](https://en.wikipedia.org/wiki/Softmax_function)
```
nalimilan, line 254: Also mention `dims` here and below? While you're at it, it would make sense to use the same wording as above, which is more explicit.

src/basicfuns.jl
```diff
@@ -259,12 +254,38 @@ $(SIGNATURES)
...
+    return r
+end
+
+function _softmax!(r, x, dims)
+    size(r) == size(x) || throw(DimensionMismatch("inconsistent array sizes"))
+    u = maximum(x; dims=dims)
+    r .= exp.(x .- u)
+    if u isa Array{eltype(r)}
+        # array can be reused
+        sum!(u, r)
```
nalimilan, line 285: Are we sure that `sum(r, dims=dims)` will always use the same eltype as `r`?

src/chainrules.jl
```diff
@@ -34,27 +34,36 @@ function ChainRulesCore.rrule(::typeof(logsumexp), x::AbstractArray{<:Real}; dim
...
-)
-    softmax!(r, x)
-    _Δx = reshape(Δx, size(r))
-    Δr = r .* (_Δx .- LinearAlgebra.dot(r, _Δx))
-    return r, Δr
+# no rules for mutating functions currently:
+# https://juliadiff.org/ChainRulesCore.jl/stable/writing_good_rules.html#Which-functions-need-rules?
+function ChainRulesCore.frule((_, Δx), ::typeof(softmax), x::AbstractArray{<:Real}; dims=:)
+    Ω = softmax(x; dims=dims)
+    ΔΩ = if dims === Colon()
```
nalimilan, line 41:
```suggestion
    ΔΩ = if dims === :
```

src/chainrules.jl
```diff
@@ -34,27 +34,36 @@ function ChainRulesCore.rrule(::typeof(logsumexp), x::AbstractArray{<:Real}; dim
...
-    rcopy = copy(reshape(r, size(x)))
-    function softmax!_pullback(r̄)
-        _r̄ = reshape(r̄, size(rcopy))
-        x̄ = ChainRulesCore.InplaceableThunk(
-            Δ -> Δ .+= rcopy .* (_r̄ .- LinearAlgebra.dot(rcopy, _r̄)),
-            ChainRulesCore.@thunk(project_x(rcopy .* (_r̄ .- LinearAlgebra.dot(rcopy, _r̄)))),
-        )
-        return ChainRulesCore.NoTangent(), ChainRulesCore.ZeroTangent(), x̄
+    function softmax_pullback(Ω̄)
+        x̄ = if dims === Colon()
```
nalimilan, line 54:
```suggestion
        x̄ = if dims === :
```

src/basicfuns.jl
devmotion: I thought it was a reasonable assumption since `eltype(r)` has to be a floating point number type. However, we could also remove this optimization or restrict it to a fixed set of types such as e.g. `Array{<:Union{Float64,Float32,Float16}}`.

src/chainrules.jl
devmotion: This is not allowed (https://github.com/JuliaStats/LogExpFunctions.jl/pull/28/checks?check_run_id=3576281866#step:6:97), therefore I'll revert the suggestion.

devmotion: And the suggestion below.

src/basicfuns.jl
devmotion: I updated the docstrings and merged the ones for the inplace method. Also I noticed that `$(SIGNATURES)` does not include the default value of the keyword arguments, and hence I wrote the signature manually.

src/basicfuns.jl
devmotion: What do you think, should it be removed completely or changed to e.g. `if eltype(r) isa Union{Float16,Float32,Float64} && u isa Array{eltype(r)}`?

src/chainrules.jl
nalimilan: Ah, I can never remember when this works or not...

docs/Manifest.toml
```diff
@@ -1,23 +1,58 @@
 # This file is machine-generated - editing it directly is not advised
```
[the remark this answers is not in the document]
devmotion: Yes, of course, it's used for instantiating the environment when building the documentation. See eg https://discourse.julialang.org/t/psa-use-a-project-for-building-your-docs/14974/7.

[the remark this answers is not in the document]
nalimilan: OK. I tend to prefer using the latest versions, so that we see what users would see. Otherwise we're likely to forget updating these (we usually forget to upgrade Documenter itself...). Anyway, that's another issue.

nalimilan (approved):

cossio:
This commit broke CI, can someone check?

devmotion:
(Ie not this PR but Julia broke CI 😛)

devmotion:
@cossio It seems the test errors are gone with the most recent Julia versions: https://github.com/JuliaStats/LogExpFunctions.jl/pull/29

PR #29, opened 2021-10-15
Define inverse functions with InverseFunctions.jl

devmotion:
This PR adds an implementation of `InverseFunctions.inverse` for `log1pexp`, `logexpm1`, `log1mexp`, `log2mexp`, `logit`, and `logistic`.

[InverseFunctions](https://github.com/JuliaMath/InverseFunctions.jl) is a lightweight package with no external dependencies that only defines the `InverseFunctions.inverse` interface and implements it for functions from base such as `exp` or `log`. In many scenarios it is useful to be able to retrieve the inverse function to a given function automatically, and InverseFunctions provides an interface for this task. In particular, it would be very useful if this would work also for `log1pexp`/`logexpm1` and `logit`/`logistic` as then the former could be used e.g. in `ParameterHandling.positive` or in GPLikelihoods as a replacement for the standard exponential transformation to the positive real numbers, without any additional code or requirement in these packages. LogExpFunctions is the only place where `inverse` can be defined without committing type piracy.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,17 +1,19 @@
 name = "LogExpFunctions"
 uuid = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 authors = ["StatsFun.jl contributors, Tamas K. Papp <tkpapp@gmail.com>"]
-version = "0.3.3"
+version = "0.3.4"
 
 [deps]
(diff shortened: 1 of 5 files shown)


tpapp:
(I am busy now, can review this at the end of Oct)

Issue #31, opened 2021-12-13
logcosh


cossio:
Currently `logcosh(x) = log(cosh(x))` is defined in NNlib like this:

```
logcosh(x) = x + softplus(-2x) - oftype(float(x), log(2))
```

It probably makes sense to have this function here, since it is a "log-exp-function" ...

devmotion:
Yes, probably makes sense to add it. We should use a numerically more stable implementation and handle `log(2)` better, eg. with something like
```julia
function logcosh(x::Real)
    abs_x = abs(x)
    return abs_x + log1pexp(-2 * abs_x) - logtwo
end
```
I'll put together a PR.

cossio:
Great thanks.

But why do you say that `abs(x) + log1pexp(-2 * abs(x))` is better than `x + softplus(-2x)`?

Do you have an example where one fails (is less accurate) and the other doesn't?

devmotion:
I'm away from my computer so I haven't tried a concrete example. It is just the natural thing to do and resembles the situation you face eg in `logaddexp`, so I assume it would be best to use the same approach here. More concretely, mathematically we have
```math
log(cosh(x)) = log(exp(-x) + exp(x)) - log(2)
```
which corresponds to
```julia
logcosh(x) = logaddexp(-x, x) - log(2)
```
I just expanded the implementation of `logaddexp` which pulls out the maximum of both values, ie. `abs(x)`. Apparently the implementation is different from the version in NNlib for negative `x`. Then the NNlib version evaluates `log1pexp` with a positive argument and it's result is added to the negative `x`. Even though we know that the result (without the `log(2)`) is larger than `abs(x)`, i.e., `-x` in this case.

cossio:
That sounds reasonable. 

Though the implementation in Keras also uses softplus: 

![image](https://user-images.githubusercontent.com/10637482/145834794-9db41e4f-59f3-4abd-95dc-f1d3e03a8869.png)
https://github.com/keras-team/keras/blob/2c48a3b38b6b6139be2da501982fd2f61d7d48fe/keras/losses.py#L1613

The only difference is that your version has an extra `abs` call, which I guess has negligible costs.

Numerically I cannot find any differences, both versions evaluate to exactly the same result for all arguments I have tried.

devmotion:
As I said, I can't test it right now - but doesn't the version without `abs` fail for `x = -Inf`?

cossio:
[...]

PR #32, opened 2021-12-14
Add `logcosh`

devmotion:
This PRs adds an implementation of `log(cosh(x))`.

Fixes https://github.com/JuliaStats/LogExpFunctions.jl/issues/31.

cc @cossio

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -94,6 +94,16 @@ logit(x::Real) = log(x / (one(x) - x))
 """
 $(SIGNATURES)
 
+Return `log(cosh(x))`.
+"""
+function logcosh(x::Real)
+    abs_x = abs(x)
+    return abs_x + log1pexp(- 2 * abs_x) - IrrationalConstants.logtwo
+end
+
+"""
+$(SIGNATURES)
+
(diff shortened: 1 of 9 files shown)


cossio:
See my comment https://github.com/JuliaStats/LogExpFunctions.jl/issues/31#issuecomment-993395336. 

But anyway I think this could be merged? It's probably accurate enough for practical purposes.

cossio:
Merge_

src/basicfuns.jl
```diff
@@ -94,6 +94,16 @@ logit(x::Real) = log(x / (one(x) - x))
 """
 $(SIGNATURES)
 
+Return `log(cosh(x))`.
```
nalimilan, line 97: Mention what's the difference with just calling `log(cosh(x))` (i.e. precision, `abs`, etc.)?

devmotion:
```suggestion
Return `log(cosh(x))`, carefully evaluated without intermediate calculation of `cosh(x)`.

The implementation ensures `logcosh(-x) = logcosh(x)`.
```

devmotion: I extended the docstring a bit :slightly_smiling_face:

nalimilan (approved):

PR #34, opened 2022-01-11
Use foldl instead of reduce for logsumexp w/o dims.

### LINKED ISSUE #33: logsumexp(::SVector) allocates
`logsumexp` allocates for static arrays, while for eg `Vector` it doesn't.
```julia
julia> VERSION
v"1.7.2-pre.0"

(@v1.7) pkg> st LogExpFunctions
      Status `~/.julia/environments/v1.7/Project.toml`
  [2ab3a3ac] LogExpFunctions v0.3.6

(@v1.7) pkg> st StaticArrays
      Status `~/.julia/environments/v1.7/Project.toml`
  [90137ffa] StaticArrays v1.3.1

julia> using StaticArrays, LogExpFunctions, BenchmarkTools

julia> x = randn(10)
10-element Vector{Float64}:
 -1.6157998430573786
  1.0318728667047257
  0.26968254689646015
  0.4903347054466403
 -1.117751634134204
 -0.4744921716401031
  0.9685578891398771
 -0.5060390401590071
  0.5209797766678386
 -0.864777496851762

julia> s = SVector(x...);

julia> @btime logsumexp($x)
  93.435 ns (0 allocations: 0 bytes)
2.5045882285138097

julia> @btime logsumexp($s)
  149.875 ns (4 allocations: 256 bytes)
2.5045882285138097
```

tpapp:
Fixes #33.

### DIFF
--- a/src/logsumexp.jl
+++ b/src/logsumexp.jl
@@ -52,11 +52,11 @@ function _logsumexp_onepass_reduce(X, ::Base.HasEltype)
     isconcretetype(T) || return _logsumexp_onepass_reduce(X, Base.EltypeUnknown())
 
     FT = float(T)
-    return reduce(_logsumexp_onepass_op, X; init=(FT(-Inf), zero(FT)))
+    return foldl(_logsumexp_onepass_op, X; init=(FT(-Inf), zero(FT)))
 end
 
 # iterables without known element type
-_logsumexp_onepass_reduce(X, ::Base.EltypeUnknown) = reduce(_logsumexp_onepass_op, X)
+_logsumexp_onepass_reduce(X, ::Base.EltypeUnknown) = foldl(_logsumexp_onepass_op, X)
 
 ## Reductions for one-pass algorithm: avoid expensive multiplications if numbers are reduced
 


devmotion:
I think I used `reduce` because of GPU compatibility. Can you check that it does not break CUDA support?

devmotion:
Another advantage of and possible motivation for `reduce` is that in contrast to `foldl` the compiler may rearrange reductions to improve performance and/or accuracy. I don't think this is exploited here currently but it is quite apparent in other simpler examples: https://discourse.julialang.org/t/reduce-vs-foldl-performance-and-precision/43231/2

tpapp:
Yes, I am aware of this and would prefer to keep `reduce` for the generic case.

How about introducing a shim 
```julia
_reduce(f, x, init) = reduce(f, x; init = init)
_reduce(f, x::StaticArray, init) = foldl(f, x; init = init)
```
and documenting this as a *workaround* that should be removed as soon as JuliaArrays/StaticArrays.jl#986 is fixed?

devmotion:
Then we would have to add a dependency on StaticArrays which seems unfortunate given how many, also central, packages depend on LogExpFunctions either directly or indirectly.

tpapp:
Closing this in favor of JuliaArrays/StaticArrays.jl#987 which fixes the issue.

PR #35, opened 2022-03-04
xexpx, xexpy

### LINKED ISSUE #15: xexpx(z)
I frequently encounter the function `xexpx(z) = z * exp(z)`, with the caveat that `xexpx(-Inf) == 0`. Thus `xexpx(z) = xlogx(exp(z))`, but computing it directly is faster because it avoids a superfluous `log` call and might prevent intermediate overflows..

Can we consider adding the definition:

```jula
xexpx(x::Real) = x == -Inf ? exp(x) : x * exp(x)
```

cossio:
Adds functions `xexpx(x)` and `xexpy(x, y)`, which compute `x * exp(x)` and `x * exp(y)`, respectively, but taking the "correct" limit when the exponent is infinite.

Close #15 .

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -44,6 +44,37 @@ function xlog1py(x::Number, y::Number)
     return iszero(x) && !isnan(y) ? zero(result) : result
 end
 
+"""
+$(SIGNATURES)
+
+Return `x * exp(x)` for `x > -Inf`, or zero if `x == -Inf`.
+
+```jldoctest
+julia> xexpx(-Inf)
+0.0
+```
+"""
+function xexpx(x::Real)
+    expx = exp(x)
+    return iszero(expx) ? expx : x * expx
+end
+
+"""
+$(SIGNATURES)
+
+Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf`.
+
+```jldoctest
+julia> xexpy(1.0, -Inf)
+0.0
+```
+"""
+function xexpy(x::Real, y::Real)
+    expy = exp(y)
+    result = x * expy
+    return iszero(expy) && !isnan(x) ? zero(result) : result
+end
+
 # The following bounds are precomputed versions of the following abstract
 # function, but the implicit interface for AbstractFloat doesn't uniformly
 # enforce that all floating point types implement nextfloat and prevfloat.
--- a/src/chainrules.jl
+++ b/src/chainrules.jl
@@ -2,6 +2,44 @@ ChainRulesCore.@scalar_rule(xlogx(x::Real), (1 + log(x),))
 ChainRulesCore.@scalar_rule(xlogy(x::Real, y::Real), (log(y), x / y,))
 ChainRulesCore.@scalar_rule(xlog1py(x::Real, y::Real), (log1p(y), x / (1 + y),))
 
+function ChainRulesCore.frule((_, Δx), ::typeof(xexpx), x::Real)
+    expx = exp(x)
+    if iszero(expx)
+        Ω = expx
+        ΔΩ = expx * Δx
+    else
+        Ω = x * expx
+        ΔΩ = (1 + x) * expx * Δx
+    end
+    return Ω, ΔΩ
+end
+
+function ChainRulesCore.rrule(::typeof(xexpx), x::Real)
+    expx = exp(x)
+    Ω = iszero(expx) ? expx : x * expx
+    function xexpx_pullback(ΔΩ)
+        Δx = iszero(expx) ? expx * ΔΩ : (1 + x) * expx * ΔΩ
+        return (ChainRulesCore.NoTangent(), Δx)
+    end
+    return Ω, xexpx_pullback
+end
+
+function ChainRulesCore.frule((_, Δx, Δy), ::typeof(xexpy), x::Real, y::Real)
+    expy = exp(y)
+    result = x * expy
+    Ω = iszero(expy) && !isnan(x) ? zero(result) : result
+    ΔΩ = expy * Δx + Ω * Δy
+    return Ω, ΔΩ
+end
+
+function ChainRulesCore.rrule(::typeof(xexpy), x::Real, y::Real)
+    expy = exp(y)
+    result = x * expy
+    Ω = iszero(expy) && !isnan(x) ? zero(result) : result
+    xexpy_pullback(ΔΩ) = (ChainRulesCore.NoTangent(), ΔΩ * expy, ΔΩ * Ω)
+    return Ω, xexpy_pullback
+end
+
 ChainRulesCore.@scalar_rule(logistic(x::Real), (Ω * (1 - Ω),))
 ChainRulesCore.@scalar_rule(logit(x::Real), (inv(x * (1 - x)),))
 ChainRulesCore.@scalar_rule(logcosh(x::Real), tanh(x))
--- a/test/basicfuns.jl
+++ b/test/basicfuns.jl
@@ -44,6 +44,37 @@
     @test iszero(xlog1py(0 + im * 0, -1 + im * Inf))
 end
 
+@testset "xexpx" begin
+    for x in (false, 0, 0.0, 0f0, -Inf, -Inf32)
+        @test (@inferred xexpx(x)) === zero(exp(x))
+    end
+    for x in (NaN16, NaN32, NaN64, Inf16, Inf32, Inf64)
+        @test (@inferred xexpx(x)) === x
+    end
+    for x in (1, true, 1.0, 1f0)
+        @test (@inferred xexpx(x)) === exp(x)
+    end
+    for a in (2, 2f0, 2.0), x in -a:a
+        @test (@inferred xexpx(x)) === x * exp(x)
+    end
+end
+
+@testset "xexpy" begin
+    for x in (0, 1, 1.0, 1f0, Inf, Inf32), y in (-Inf, -Inf32)
+        @test (@inferred xexpy(x, y)) === zero(x * exp(y))
+    end
+    for x in (0, 1, 1.0, 1f0, Inf, Inf32, -Inf, -Inf32, NaN, NaN32), nan in (NaN, NaN32)
+        @test (@inferred xexpy(x, nan)) === oftype(x * exp(nan), NaN)
+        @test (@inferred xexpy(nan, x)) === oftype(nan * exp(x), NaN)
+    end
+    for x in (2, -2f0, 2.0), y in (1, -1f0, 1.0)
+        @test (@inferred xexpy(x, y)) ≈ x * exp(y)
+    end
+    for x in (randn(), randn(Float32))
+        @test xexpy(x, x) ≈ xexpx(x)
+    end
+end
+
 @testset "logistic & logit" begin
     @test logistic(2) ≈ 1.0 / (1.0 + exp(-2.0))
     @test logistic(-750.0) === 0.0
--- a/test/chainrules.jl
+++ b/test/chainrules.jl
@@ -13,6 +13,27 @@
         end
     end
 
+    @testset "xexpx" begin
+        # regular branch
+        test_scalar(xexpx, randn())
+        # special cases (manually since FiniteDifferences/ChainRulesTestUtils fails at -Inf)
+        @test @inferred(frule((NoTangent(), rand()), xexpx, -Inf)) === (0.0, 0.0)
+        Ω, back = @inferred(rrule(xexpx, -Inf))
+        @test Ω === 0.0
+        @test back(rand()) === (NoTangent(), 0.0)
+    end
+
+    @testset "xexpy" begin
+        # regular branch
+        test_frule(xexpy, randn(), randn())
+        test_rrule(xexpy, randn(), randn())
+        # special cases (manually since FiniteDifferences/ChainRulesTestUtils fails at -Inf)
+        @test @inferred(frule((NoTangent(), rand(), rand()), xexpy, x, -Inf)) === (0.0, 0.0)
+        Ω, back = @inferred(rrule(xexpy, x, -Inf))
+        @test Ω === 0.0
+        @test back(rand()) === (NoTangent(), 0.0, 0.0)
+    end
+
     test_frule(logit, x)
     test_rrule(logit, x)
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LogExpFunctions"
 uuid = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 authors = ["StatsFun.jl contributors, Tamas K. Papp <tkpapp@gmail.com>"]
-version = "0.3.6"
+version = "0.3.7"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -10,6 +10,8 @@ LogExpFunctions supports [`InverseFunctions.inverse`](https://github.com/JuliaMa
 xlogx
 xlogy
 xlog1py
+xexpx
+xexpy
 logistic
 logit
 logcosh
--- a/src/LogExpFunctions.jl
+++ b/src/LogExpFunctions.jl
@@ -9,7 +9,7 @@ import InverseFunctions
 import IrrationalConstants
 import LinearAlgebra
 
-export xlogx, xlogy, xlog1py, logistic, logit, log1psq, log1pexp, log1mexp, log2mexp, logexpm1,
+export xlogx, xlogy, xlog1py, xexpx, xexpy, logistic, logit, log1psq, log1pexp, log1mexp, log2mexp, logexpm1,
     softplus, invsoftplus, log1pmx, logmxp1, logaddexp, logsubexp, logsumexp, softmax,
     softmax!, logcosh
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,6 @@
 using LogExpFunctions
 using ChainRulesTestUtils
+using ChainRulesCore
 using ChangesOfVariables
 using InverseFunctions
 using OffsetArrays


test/basicfuns.jl
```diff
@@ -44,6 +44,37 @@
     @test iszero(xlog1py(0 + im * 0, -1 + im * Inf))
 end
 
+@testset "xexpx" begin
+    @test iszero(xexpx(-Inf))
+    @test iszero(xexpx(0))
+    @test isnan(xexpx(NaN))
```
devmotion, line 50: Maybe also test `xexpx(Inf)`.

test/basicfuns.jl
```diff
@@ -44,6 +44,37 @@
...
+
+    @test xexpx(1) ≈ exp(1)
+    @test xexpx(2) ≈ 2exp(2)
+
+    @inferred xexpx(0)
+    @inferred xexpx(0.)
+    @inferred xexpx(0f0)
```
devmotion, line 57: Can you check the type as well?

test/basicfuns.jl
```diff
@@ -44,6 +44,37 @@
...
+end
+
+@testset "xexpy" begin
+    @test iszero(xexpy(Inf, -Inf))
```
devmotion, line 61: Mathematically the limit does not exist.

test/basicfuns.jl
```diff
@@ -44,6 +44,37 @@
...
+
+    @test isnan(xexpy(NaN, -Inf))
+    @test isnan(xexpy(NaN, 1))
+    @test isnan(xexpy(0, NaN))
+
+    @inferred xexpy(0,-Inf)
+    @inferred xexpy(1,1)
+    @inferred xexpy(1.,-Inf)
```
devmotion, line 69: Again, check the actual types as well?

devmotion (commented):
Thanks!

Can you add and test ChainRules derivatives as well?

test/basicfuns.jl
cossio: Yeah but having `xexpy(Inf, -Inf) == 0` is the point of having `xexpy` at all (just like `xlogy`)

devmotion: Sure, I agree. I just think it is weird/incorrect to talk about taking limits in the docstring and one should rather explain the special case, i.e., how the implementation differs from `x * exp(-y)`.

test/basicfuns.jl
tpapp: Or maybe combine with value tests, eg
```julia
@test @inferred(xexpx(0.0)) === 0.0
```
etc.

src/basicfuns.jl
```diff
@@ -44,6 +44,31 @@ function xlog1py(x::Number, y::Number)
...
+0.0
+```
+"""
+xexpx(x::Real) = isfinite(x) ? x * exp(x) : exp(x)
+
+"""
+$(SIGNATURES)
+
+Return `x * exp(y)` for `y > -Inf`, handling `y = -Inf` by taking the downward limit.
+
```
tpapp, line 63: I find this a bit confusing (how the limit is defined). I think the manual should just say that it is `x * exp(y)` except when ..., and enumerate the special cases and their evaluation rules explicitly.

src/basicfuns.jl
```diff
@@ -44,6 +44,31 @@ function xlog1py(x::Number, y::Number)
     return iszero(x) && !isnan(y) ? zero(result) : result
 end
 
+"""
+$(SIGNATURES)
+
+Return `x * exp(x)` for `x > -Inf`, handling `x = -Inf` by taking the downward limit.
+
```
tpapp, line 51: Instead of using limits, I would suggest just describing the special case explicitly (`exp(-Inf) == 0.0`).

test/basicfuns.jl
tpapp: Or, similarly, combine with value tests.

test/basicfuns.jl
cossio: What do you mean?

src/basicfuns.jl
cossio: Yes that's better I updated the docstring. Thanks.

src/basicfuns.jl
cossio: Also updated this docstring

test/basicfuns.jl
cossio: I updated the docstrings as suggested by @tpapp to make this point clearer.

test/basicfuns.jl
cossio: Yep

cossio:
I added the ChainRules and addressed your comments @devmotion @tpapp . Care for another look?

src/basicfuns.jl
```diff
@@ -44,6 +44,31 @@ function xlog1py(x::Number, y::Number)
...
+$(SIGNATURES)
+
+Return `x * exp(x)` for `x > -Inf`, or zero if `x == -Inf`.
+
+```jldoctest
+julia> xexpx(-Inf)
+0.0
+```
+"""
+xexpx(x::Real) = isfinite(x) ? x * exp(x) : exp(x)
```
devmotion, line 57: The check is a bit simpler and more direct (since currently the second bramch is used not only for `-Inf`) if you write

```suggestion
function xexpx(x::Real)
    expx = exp(x)
    iszero(expx) ? expx : x * expx
end
```

src/basicfuns.jl
```diff
@@ -44,6 +44,31 @@ function xlog1py(x::Number, y::Number)
...
+$(SIGNATURES)
+
+Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf`.
+
+```jldoctest
+julia> xexpy(1.0, -Inf)
+0.0
+```
+"""
+xexpy(x::T, y::T) where {T<:Real} = isnan(x) || isfinite(y) ? x * exp(y) : exp(y)
```
devmotion, line 69: Similarly, here one could use

```suggestion
function xexpy(x::T, y::T) where {T<:Real}
    expy = exp(y)
    isnan(x) || iszero(expy) ? x * expy : expy
end
```

src/chainrules.jl
```diff
@@ -2,6 +2,15 @@ ChainRulesCore.@scalar_rule(xlogx(x::Real), (1 + log(x),))
 ChainRulesCore.@scalar_rule(xlogy(x::Real, y::Real), (log(y), x / y,))
 ChainRulesCore.@scalar_rule(xlog1py(x::Real, y::Real), (log1p(y), x / (1 + y),))
 
+ChainRulesCore.@scalar_rule(
+    xexpx(x::Real),
+    (iszero(x) ? one(Ω) : isfinite(x) ? Ω / x * (1 + x) : Ω,)
```
devmotion, line 7: This is simply
```suggestion
    (xexpy(x + 1, x),)
```
If one wants to reuse part of the forward pass (`exp(x)`) one should implement `frule` and `rrule` explicitly but not divide the result by `x`. It's an optimization which we don't perform for `xlogy` etc. either, so I think `xexpy` is fine in this PR.

src/chainrules.jl
```diff
@@ -2,6 +2,15 @@ ChainRulesCore.@scalar_rule(xlogx(x::Real), (1 + log(x),))
...
+)
+ChainRulesCore.@scalar_rule(
+    xexpy(x::Real, y::Real),
+    (iszero(x) ? oftype(Ω, exp(y)) : Ω / x, Ω)
```
devmotion, line 11: Similarly,

```suggestion
    (exp(y), Ω)
```

test/basicfuns.jl
```diff
@@ -44,6 +44,30 @@
...
+    @test 2exp(2) ≈ @inferred xexpx(2)
+end
+
+@testset "xexpy" begin
+    @test iszero(@inferred xexpy(Inf, -Inf))
+    @test isnan(@inferred xexpy(NaN, -Inf))
+    @test isnan(@inferred xexpy(NaN, 1))
+    @test isnan(@inferred xexpy(0, NaN))
+    @test iszero(@inferred xexpy(0, -Inf))
+    @test ℯ ≈ @inferred xexpy(1, 1)
```
devmotion, line 63: Maybe rather use
```suggestion
    @test e ≈ @inferred xexpy(1, 1)
```
or
```suggestion
    @test exp(1) ≈ @inferred xexpy(1, 1)
```

devmotion: Ah wait, it's covered by the loop below already:
```suggestion
```

test/basicfuns.jl
```diff
@@ -44,6 +44,30 @@
...
+    @test iszero(@inferred xexpy(1., -Inf))
+    @test 2exp(3) ≈ @inferred xexpy(2, 3)
```
devmotion, line 65: Same here
```suggestion
```

test/chainrules.jl
```diff
@@ -13,6 +13,15 @@
         end
     end
 
+    for x in -2.0:2.0
+        test_frule(xexpx, x)
+        test_rrule(xexpx, x)
```
devmotion, line 18:
```suggestion
        test_scalar(xexpx, x)
```

test/chainrules.jl
```diff
@@ -13,6 +13,15 @@
...
+        for y in -2.0:2.0
+            test_frule(xexpy, x, y)
+            test_rrule(xexpy, x, y)
```
devmotion, line 21:
```suggestion
            test_scalar(xexpy, x, y)
```

src/basicfuns.jl
tpapp: I think the original code matches the intention very well though, so I would be OK with it as is.

src/basicfuns.jl
tpapp: Perfect.

src/basicfuns.jl
tpapp: Same here, I am OK with the original code.

tpapp (changes requested):
As I said in the comments, I am fine with the branch structure as is, so it is up to you. For all other minor changes I agree with @devmotion's comments (marked them by a :+1:).

Once this is done, this PR should be ready to merge. Thanks for doing it!

src/chainrules.jl
cossio: What's wrong with `Ω / x`?

test/basicfuns.jl
cossio: That's right, thanks!

test/chainrules.jl
cossio: Didn't know about `test_scalar`. But that should be done for all the other rules too ...

test/chainrules.jl
cossio: There is no `test_scalar(xexpy, x, y)` with multiple args :(

src/basicfuns.jl
cossio: Honestly I also prefer the original version :)

cossio:
I addressed the comments. Only thing I don't understand is the objection to using `Ω / x` in the chain rule. I'd rather argue it is worth doing this optimization also for `xlogx`, etc. (in another PR)

src/basicfuns.jl
devmotion: I still think `iszero` is clearer since it focuses on the one special case we want to handle (similar to the `iszero` in eg `xlogx`). `isfinite` catches other cases as well and is a more complicated check, for `AbstractFloat` (https://github.com/JuliaLang/julia/blob/fe1264a5c980f9a17be715f402fce592c7f6f35c/base/float.jl#L499) and in particular generic `Real` (https://github.com/JuliaLang/julia/blob/fe1264a5c980f9a17be715f402fce592c7f6f35c/base/float.jl#L500).

devmotion (changes requested):
> Only thing I don't understand is the objection to using Ω / x in the chain rule @devmotion . I'd rather argue it is worth doing this optimization also for xlogx, etc. (in another PR)

I guess maybe I didn't express myself clearly enough: I'm not against optimizing the forward and reverse mode rules but I think computing `Ω / x` is the wrong thing to do here. Instead one should implement `frule` and `rrule` manually and reuse the result of `exp(x)` and `exp(y)` from the computation of the primal.

But I think it's fine to do that in a separate PR and use the unoptimized version initially.

cossio:
> but I think computing `Ω / x` is the wrong thing to do here

Why?

devmotion:
Because it is an unnecessary computation that you can avoid by just reusing `exp(x)`/`exp(y)`, and it requires you to handle cases such as `iszero(x)`. Hence IMO we should rather use the unoptimized version or implement `frule`/`rrule`.

cossio:
I implemented the explicit `frule` and `rrule` as suggested by @devmotion . I also went ahead and used your suggested branches in the primal definitions since this looked more consistent with how the chain rules looked when reusing `exp(x)` etc.

I also bumped the version. Can we tag a release after merging?

Care to take another look? @devmotion @tpapp

src/chainrules.jl
```diff
@@ -2,6 +2,48 @@ ChainRulesCore.@scalar_rule(xlogx(x::Real), (1 + log(x),))
...
+        Ω = x * e
+        ΔΩ = (1 + x) * e * Δx
+    end
+    return Ω, ΔΩ
+end
+
+function ChainRulesCore.rrule(::typeof(xexpx), x::Real)
+    e = exp(x)
+    Ω = iszero(e) ? e : x * e
+    xexpx_pullback(Δ) = (ChainRulesCore.NoTangent(), iszero(e) ? e * Δ : (1 + x) * e * Δ)
```
devmotion, line 20: I'd suggest breaking up the pullback to increase readability and to rename `e`. E.g.,
```suggestion
    expx = exp(x)
    Ω = iszero(expx) ? expx : x * expx
    function xexpx_pullback(ΔΩ)
        Δx = iszero(expx) ? expx * ΔΩ : (1 + x) * expx * ΔΩ
        return ChainRulesCore.NoTangent(), Δx
    end
```

src/chainrules.jl
```diff
@@ -2,6 +2,48 @@ ChainRulesCore.@scalar_rule(xlogx(x::Real), (1 + log(x),))
...
+
+function ChainRulesCore.frule((_, Δx, Δy), ::typeof(xexpy), x::T, y::T) where {T<:Real}
+    e = exp(y)
+    if iszero(e) && !isnan(x)
+        Ω = e
+        ΔΩ = e * Δx + zero(Ω * Δy)
+    else
+        Ω = x * e
+        ΔΩ = e * Δx + Ω * Δy
+    end
```
devmotion, line 32: `e * Δx + zero(Ω * Δy)` is the same as `e * Δx + Ω * Δy`, isn't it? Similar to above, I'd also suggest using more descriptive variable names, e.g.
```suggestion
    expy = exp(y)
    Ω = iszero(expy) && !isnan(x) ? expy : x * expy
    ΔΩ = expy * Δx + Ω * Δy
```

src/chainrules.jl
```diff
@@ -2,6 +2,48 @@ ChainRulesCore.@scalar_rule(xlogx(x::Real), (1 + log(x),))
...
+end
+
+function ChainRulesCore.rrule(::typeof(xexpy), x::T, y::T) where {T<:Real}
+    e = exp(y)
+    Ω = iszero(e) && !isnan(x) ? e : x * e
+    xexpy_pullback(Δ) = if iszero(e) && !isnan(x)
+        (ChainRulesCore.NoTangent(), e * Δ, zero(Ω * Δ))
+    else
+        (ChainRulesCore.NoTangent(), e * Δ, Ω * Δ)
+    end
```
devmotion, line 43: Again, I'd suggest to simplify and improve readability by using
```suggestion
    expy = exp(y)
    Ω = iszero(expy) && !isnan(x) ? expy : x * expy
    function xexpy_pullback(ΔΩ)
        Δx = expy * ΔΩ
        Δy = Ω * ΔΩ
        return ChainRulesCore.NoTangent(), Δx, Δy
    end
```

test/basicfuns.jl
```diff
@@ -44,6 +44,34 @@
     @test iszero(xlog1py(0 + im * 0, -1 + im * Inf))
 end
 
+@testset "xexpx" begin
+    for z in (0, 0.0, 0f0, false, -Inf)
+        @test iszero(@inferred xexpx(z))
+    end
+    for nan in (NaN16, NaN32, NaN64)
+        @test isnan(@inferred xexpx(nan))
```
devmotion, line 52: Could also check that the type is as expected:
```suggestion
        @test @inferred(xexpx(nan)) === nan
```

test/basicfuns.jl
```diff
@@ -44,6 +44,34 @@
...
+    end
+    for inf in (Inf16, Inf32, Inf64)
+        @test inf === @inferred xexpx(inf)
```
devmotion, line 55: IMO it's a bit more natural to flip LHS and RHS, i.e. use
```suggestion
        @test @inferred(xexpx(inf)) === inf
```
Not a big deal though :slightly_smiling_face:

test/basicfuns.jl
```diff
@@ -44,6 +44,34 @@
...
+    end
+    for u in (1, true, 1.0, 1f0)
+        @test exp(u) === @inferred xexpx(u)
```
devmotion, line 58:
```suggestion
        @test @inferred(xexpx(u)) === exp(u)
```

test/basicfuns.jl
```diff
@@ -44,6 +44,34 @@
...
+    end
+    @test 2exp(2) ≈ @inferred xexpx(2)
```
devmotion, line 60:
```suggestion
    @test @inferred(xexpx(2)) ≈ 2 * exp(2)
```

devmotion: Alternatively, one could also test a random number:
```suggestion
    for x in (randn(), randn(Float32))
        @test @inferred(xexpx(x)) isa typeof(x)
        @test xexpx(x) ≈ x * exp(x)
    end
```

test/basicfuns.jl
```diff
@@ -44,6 +44,34 @@
...
+end
+
+@testset "xexpy" begin
+    @test iszero(@inferred xexpy(Inf, -Inf))
+    @test isnan(@inferred xexpy(NaN, -Inf))
+    @test isnan(@inferred xexpy(NaN, 1))
+    @test isnan(@inferred xexpy(0, NaN))
+    @test iszero(@inferred xexpy(0, -Inf))
+    @test iszero(@inferred xexpy(1., -Inf))
```
devmotion, line 69: I guess it might be clearer to reorder the tests. Additionally, I suggest adding tests with other types, e.g.,
```suggestion
    for x in (-Inf, -Inf32)
        @test @inferred(xexpy(0, x)) === zero(x)
        @test @inferred(xexpy(one(x), x)) === zero(x)
        @test @inferred(xexpy(x, x)) === zero(x)
        @test @inferred(xexpy(-x, x)) === zero(x)
    end

    for x in (NaN, NaN32)
        @test @inferred(xexpy(x, oftype(x, -Inf))) === x
        @test @inferred(xexpy(x, 1)) === x
        @test @inferred(xexpy(x, Inf32)) === x
        @test @inferred(xexpy(-Inf32, x)) === x
        @test @inferred(xexpy(0, x)) === x
        @test @inferred(xexpy(oftype(x, Inf), x)) === x
        @test @inferred(xexpy(x, x)) === x
    end
```

test/basicfuns.jl
```diff
@@ -44,6 +44,34 @@
...
+    for x = -3:3, y = -3:3
+        @test x * exp(y) ≈ @inferred xexpy(x, y)
+    end
```
devmotion, line 72: Again, maybe it's sufficient to test one/two random numbers, e.g.
```suggestion
    for x in (randn(), randn(Float32)), y in (randn(), randn(Float32))
        correct = x * exp(y)
        @test @inferred(xexpy(x, y)) isa typeof(correct)
        @test xexpy(x, y) ≈ correct
        @test xexpy(x, x) ≈ xexpx(x)
        @test xexpy(y, y) ≈ xexpx(y)
    end
```

test/chainrules.jl
```diff
@@ -13,6 +13,14 @@
         end
     end
 
+    for x in -2.0:2.0
+        test_scalar(xexpx, x)
+        for y in -2.0:2.0
+            test_frule(xexpy, x, y)
+            test_rrule(xexpy, x, y)
+        end
+    end
```
devmotion, line 22: It should be sufficient to check a single random pair of `x` and `y` since there are no special branches among the tested values. Additionally, it would be good to check the special cases, i.e., `xexpx` with `x = -Inf` and `xexpy` with `y = -Inf`. I don't think this is possible with ChainRulesTestUtils/FiniteDifferences though since the non-finite values will mess with finite differencing. So instead probably one just has to check the output of `frule` and `rrule` manually.
```suggestion
    x = randn()
    y = randn()
    # Test regular branch
    test_scalar(xexpx, x)
    test_frule(xexpy, x, y)
    test_rrule(xexpy, x, y)

    # Test special cases:
    # manual checks since FiniteDifferences/ChainRulesTestUtils can't differentiate around `-Inf`
    @test @inferred(frule((NoTangent(), randn()), xexpx, -Inf)) === (0.0, 0.0)
    Ω, back = @inferred(rrule(xexpx, -Inf))
    @test Ω === 0.0
    @test back(randn()) === (NoTangent(), 0.0)
    @test @inferred(frule((NoTangent(), randn(), randn()), xexpy, x, -Inf)) === (0.0, 0.0)
    Ω, back = @inferred(rrule(xexpy, x, -Inf))
    @test Ω === 0.0
    @test back(randn()) === (NoTangent(), 0.0, 0.0)
```

devmotion (changes requested):
I think it looks good. I made some suggestions for how the code could be made more readable and simpler (in particular the CR derivatives), and how the tests could be extended.

cossio:
Incorporated most of your comments (with minor modifications) @devmotion

src/basicfuns.jl
```diff
@@ -44,6 +44,36 @@ function xlog1py(x::Number, y::Number)
...
+Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf`.
+
+```jldoctest
+julia> xexpy(1.0, -Inf)
+0.0
+```
+"""
+function xexpy(x::Real, y::Real)
+    expy = exp(y)
+    return iszero(expy) && !isnan(x) ? zero(x * expy) : x * expy
```
devmotion, line 74: What's the reason for changing it to `zero(x * expy)`? To avoid the promotion? Similar to `xlogy` etc. maybe one could even use
```suggestion
    result = x * expy
    return iszero(expy) && !isnan(x) ? zero(result) : result
```
Possibly this is faster if branch prediction fails.

src/chainrules.jl
```diff
@@ -2,6 +2,42 @@ ChainRulesCore.@scalar_rule(xlogx(x::Real), (1 + log(x),))
...
+    function xexpx_pullback(ΔΩ)
+        Δx = iszero(expx) ? expx * ΔΩ : (1 + x) * expx * ΔΩ
+        return (ChainRulesCore.NoTangent(), Δx)
+    end
+    return Ω, xexpx_pullback
+end
+
+function ChainRulesCore.frule((_, Δx, Δy), ::typeof(xexpy), x::Real, y::Real)
+    expy = exp(y)
+    Ω = iszero(expy) && !isnan(x) ? zero(x * expy) : x * expy
```
devmotion, line 29: Same here, possibly use
```suggestion
    result = x * expy
    Ω = iszero(expy) && !isnan(x) ? zero(result) : result
```

src/chainrules.jl
```diff
@@ -2,6 +2,42 @@ ChainRulesCore.@scalar_rule(xlogx(x::Real), (1 + log(x),))
...
+    ΔΩ = expy * Δx + Ω * Δy
+    return Ω, ΔΩ
+end
+
+function ChainRulesCore.rrule(::typeof(xexpy), x::Real, y::Real)
+    expy = exp(y)
+    Ω = iszero(expy) && !isnan(x) ? zero(x * expy) : x * expy
```
devmotion, line 36: Same here, maybe
```suggestion
    result = x * expy
    Ω = iszero(expy) && !isnan(x) ? zero(result) : result
```

test/chainrules.jl
```diff
@@ -13,6 +13,27 @@
         end
     end
 
+    @testset "xexpx" begin
+        # regular branch
+        test_scalar(xexpx, randn())
+        # special cases (manually since FiniteDifferences/ChainRulesTestUtils fails at -Inf)
+        @test @inferred(frule((NoTangent(), 1), xexpx, -Inf)) === (0.0, 0.0)
```
devmotion, line 20: I think it's better to use `randn` to ensure that the result is independent of the tangent:
```suggestion
        @test @inferred(frule((NoTangent(), randn()), xexpx, -Inf)) === (0.0, 0.0)
```

test/chainrules.jl
```diff
@@ -13,6 +13,27 @@
...
+        Ω, back = @inferred(rrule(xexpx, -Inf))
+        @test Ω === 0.0
+        @test back(1) === (NoTangent(), 0.0)
```
devmotion, line 23: Same here:
```suggestion
        @test back(randn()) === (NoTangent(), 0.0)
```

test/chainrules.jl
```diff
@@ -13,6 +13,27 @@
...
+    end
+
+    @testset "xexpy" begin
+        # regular branch
+        test_frule(xexpy, randn(), randn())
+        test_rrule(xexpy, randn(), randn())
+        # special cases (manually since FiniteDifferences/ChainRulesTestUtils fails at -Inf)
+        @test @inferred(frule((NoTangent(), 1, 1), xexpy, x, -Inf)) === (0.0, 0.0)
```
devmotion, line 31:
```suggestion
        @test @inferred(frule((NoTangent(), randn(), randn()), xexpy, x, -Inf)) === (0.0, 0.0)
```

src/basicfuns.jl
cossio: I figure `zero(x * expy)` was more readable. Since ternary operator short-circuits the product `x * expy` will not be evaluated more than once anyway.

src/chainrules.jl
cossio: See comment above.

test/chainrules.jl
cossio: That can fail because the tangent can be `-0.0` or `0.0`, depending on the sign output by `randn()`.

cossio: I can use `rand()` though.

src/basicfuns.jl
devmotion: I'm pretty sure that the ternary operator does not short-circuit, and only logical operators such as `&&` and `||` do. In contrast to `ifelse` both branches don't have to be evaluated (there per se it's a regular function call and hence arguments have to be evaluated), but there's no guarantee that the branch prediction of the CPU always guesses the correct branch and only one of the two branches is evaluated.

cossio: Oh you are right, ternary operator does not short-circuit. But I'd guess common sub-expressions would be extracted from the branch by the compiler. Anyway the `result = ...` trick is used also in `xlogx`, etc., so I went with that.

devmotion (approved):
Thanks for the PR and for addressing all our comments :slightly_smiling_face: 

I think the PR is ready to be merged and released but I want to wait if @tpapp has any additional comments.

tpapp (approved):
LGTM. Thanks again!

PR #37, opened 2022-03-07
log1pexp

### LINKED ISSUE #13: log1pexp(x) when x < -37
I was reading R's implementation of `log1pexp` and they include one additional branch (see https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf, equation 10). 

Currently StatsFuns does this:

https://github.com/JuliaStats/StatsFuns.jl/blob/ae40bc1c6ee63a5624ba6cd168da5bf43f473e59/src/basicfuns.jl#L66

Whereas including the additional branch it would be:

```julia
log1pexp_v2(x::AbstractFloat) = x ≤ -37. ? exp(x) : x ≤ 18. ? log1p(exp(x)) : x ≤ 33.3 ? x + exp(-x) : x
```

Apparently including the `x < -37.` check reduces the function's computation time by 25% (because it saves the call to `log1p` I guess). But I am no expert doing this kind of performance tests so please can someone else more knowledgeable check this? Here is the test I did:

```julia
using BenchmarkTools
@btime for x = -100. : 1. : 100., _ = 1 : 100
    log1pexp_v2(x)
end   #  310.471 μs (0 allocations: 0 bytes)
@btime for x = -100. : 1. : 100., _ = 1 : 100
    log1pexp(x)
end  #   423.671 μs (0 allocations: 0 bytes)
```

Both `log1pexp_v2` and `log1pexp` give numerically identical results.

If this is worth it I can prepare a PR. 

**Update:** Actually it seems much of the time is due to the `oftype(exp(-x), x)` bit in the final branch. I think specializing on `AbstractFloat` would take care of that. Even after this change the `x ≤ -37.` check is faster.

cossio:
Based on
https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf

Close #13

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -168,7 +168,12 @@ function log1pexp(x::Real)
 end
 
 # returns a threshold such that log1pexp(x) ≈ x for x > threshold
-@generated _log1pexp_threshold(::T) where {T<:Real} = -log(expm1(eps(float(T))/2))
+@generated function _log1pexp_threshold(::T) where {T<:Real}
+    F = float(T)
+    ϵ = big(eps(F)) / 2
+    thresh = -log(expm1(ϵ))
+    return convert(F, thresh)
+end
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -202,7 +202,7 @@ should optimize away all computations done here.
 end
 
 #= For Float64, Float32 we can hard-code the thresholds to make absolutely sure they are not
-recompued each time. Also, _log1pexp_thresholds is not completely elided by the
+recomputed each time. Also, _log1pexp_thresholds is not completely elided by the
 compiler in Julia 1.0 / 1.6 which LogExpFunctions intends to support.
 For Float64 we use the same exact thresholds given by Maechler 2012,
 since these were the ones used before _log1pexp_thresholds was introduced (which outputs
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -185,7 +185,7 @@ where the tolerances of the approximations ≈ are on the order of eps(T)
     x0 = log(ϵ / 2)
     x1 = -log(ϵ) / 2
     x2 = -log(ϵ/2) - log(-log(ϵ/2)) - log(-log(ϵ/2)) / log(ϵ/2) # root of e^-x == x * ϵ/2
-    return F.((x0, x1, x2))
+    return (F(x0), F(x1), F(x2))
 end
 
 """
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -158,11 +158,11 @@ See:
 """
 function log1pexp(x::Real)
     x0, x1, x2 = _log1pexp_thresholds(x)
-    if x ≤ x0
+    if x < x0
         return exp(x)
-    elseif x ≤ x1
+    elseif x < x1
         return log1p(exp(x))
-    elseif x ≤ x2
+    elseif x < x2
         return x + exp(-x)
     else
         return float(x)
@@ -184,7 +184,7 @@ where the tolerances of the approximations ≈ are on the order of eps(T)
     ϵ = big(eps(F))
     x0 = log(ϵ / 2)
     x1 = -log(ϵ) / 2
-    x2 = -x0 - log(-x0) * (1 + 1 / x0) # root of e^-x == x * ϵ/2
+    x2 = -x0 - log(-x0) * (1 + 1 / x0) # ≈ root of e^-x == x * ϵ/2
     return (F(x0), F(x1), F(x2))
 end
 
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -165,7 +165,7 @@ function log1pexp(x::Real)
     elseif x ≤ x2
         return x + exp(-x)
     else
-        return oftype(x0, x)
+        return float(x)
     end
 end
 
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -157,42 +157,37 @@ See:
  * Martin Maechler (2012) [“Accurately Computing log(1 − exp(− |a|))”](http://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf)
 """
 function log1pexp(x::Real)
-    thresh_lin = _log1pexp_threshold_x2(x)
-    if x > thresh_lin
-        return oftype(thresh_lin, x)
-    elseif x > _log1pexp_threshold_x1(x)
-        return x + exp(-x)
-    elseif x ≤ _log1pexp_threshold_exp(x)
+    x0, x1, x2 = _log1pexp_thresholds(x)
+    if x ≤ x0
         return exp(x)
+    elseif x ≤ x1
+        return log1p(exp(x))
+    elseif x ≤ x2
+        return x + exp(-x)
     else
-    	return log1p(exp(x))
-  	end
+        return oftype(x0, x)
+    end
 end
 
-# returns a threshold such that log1pexp(x) ≈ x for x > threshold
-@generated function _log1pexp_threshold_x2(::T) where {T<:Real}
+#=
+returns a thresholds x0, x1, x2 such that:
+    * log1pexp(x) ≈ exp(x) for x ≤ x0
+    * log1pexp(x) ≈ log1p(exp(x)) for x0 < x ≤ x1
+    * log1pexp(x) ≈ x + exp(-x) for x1 < x ≤ x2
+    * log1pexp(x) ≈ x for x > x2
+=#
+@generated function _log1pexp_thresholds(::T) where {T<:Real}
     F = float(T)
-    thresh = -log(expm1(big(eps(F)) / 2))
-    return convert(F, max(thresh, log(big(ℯ) - 1)))
-end
-# tighter thresholds can be obtained numerically. Since this is the fastest branch,
-# we can do it here for common types.
-_log1pexp_threshold_x2(::Float64) = 33.3e0
-_log1pexp_threshold_x2(::Float32) = 15.9f0
-
-# returns a threshold such that log1pexp(x) ≈ x + exp(-x) for x > threshold
-@generated function _log1pexp_threshold_x1(::T) where {T<:Real}
-    F = float(T)
-    thresh = -log(big(eps(F))) / 2
-    return convert(F, max(thresh, log(big(ℯ) - 1)))
+    ϵ = big(eps(F))
+    x0 = log(ϵ / 2)
+    x1 = -log(ϵ) / 2
+    x2 = -log(expm1(ϵ / 2))
+    return F.((x0, x1, x2))
 end
 
-# returns a threshold such that log1pexp(x) ≈ exp(x) for x < threshold
-@generated function _log1pexp_threshold_exp(::T) where {T<:Real}
-	F = float(T)
-	thresh = log(big(eps(F)) / 2)
-	return convert(F, thresh)
-end
+# tighter thresholds can be obtained numerically for common types
+_log1pexp_thresholds(::Float64) = (-37.0, 18.0, 33.0)
+_log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 15.9458f0)
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -202,12 +202,15 @@ should optimize away all computations done here.
 end
 
 #= For Float64, Float32 we can hard-code the thresholds to make absolutely sure they are not
-recompued each time. For Float64 we use the same exact thresholds given by Maechler 2012,
+recompued each time. Also, _log1pexp_thresholds is not completely elided by the
+compiler in Julia 1.0 / 1.6 which LogExpFunctions intends to support.
+For Float64 we use the same exact thresholds given by Maechler 2012,
 since these were the ones used before _log1pexp_thresholds was introduced (which outputs
 close but not identical thresholds), to reduce any (small) risk of breakage.
-For Float32 we use truncated versions of the output of _log1pexp_thresholds.  =#
+For Float32 and Float16 we use truncated versions of the output of _log1pexp_thresholds. =#
 @inline _log1pexp_thresholds(::Float64) = (-37e0, 18e0, 33e0)
-@inline _log1pexp_thresholds(::Float32) = (-17f0, 9f0, 16f0)
+@inline _log1pexp_thresholds(::Float32) = (-17f0, 8f0, 14f0)
+@inline _log1pexp_thresholds(::Float16) = (Float16(-7.6), Float16(3.5), Float16(5.9))
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -170,9 +170,8 @@ function log1pexp(_x::Real)
     end
 end
 
-#= The precision of BigFloat cannot be computed from the type only.
-Since computing the dynamically thresholds is slow, prefer to use
-a slower but accurate implementation in this case. =#
+#= The precision of BigFloat cannot be computed from the type only and computing
+thresholds is slow. Therefore prefer version without thresholds in this case. =#
 function log1pexp(x::BigFloat)
     if x > 0
         return x + log1p(exp(-x))
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -156,7 +156,8 @@ transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`log
 See:
  * Martin Maechler (2012) [“Accurately Computing log(1 − exp(− |a|))”](http://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf)
 """
-function log1pexp(x::Real)
+function log1pexp(_x::Real)
+    x = float(_x)
     x0, x1, x2 = _log1pexp_thresholds(x)
     if x < x0
         return exp(x)
@@ -165,7 +166,7 @@ function log1pexp(x::Real)
     elseif x < x2
         return x + exp(-x)
     else
-        return float(x)
+        return x
     end
 end
 
@@ -180,12 +181,11 @@ returns a thresholds x0, x1, x2 such that:
 where the tolerances of the approximations ≈ are on the order of eps(T)
 =#
 @inline @generated function _log1pexp_thresholds(::T) where {T<:Real}
-    F = float(T)
-    ϵ = big(eps(F))
+    ϵ = big(eps(T))
     x0 = log(ϵ / 2)
     x1 = -log(ϵ) / 2
     x2 = -x0 - log(-x0) * (1 + 1 / x0) # ≈ root of e^-x == x * ϵ/2
-    return (F(x0), F(x1), F(x2))
+    return (T(x0), T(x1), T(x2))
 end
 
 """
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -181,21 +181,9 @@ returns a thresholds x0, x1, x2 such that:
     ϵ = big(eps(F))
     x0 = log(ϵ / 2)
     x1 = -log(ϵ) / 2
-    x2 = -log(expm1(ϵ / 2))
+    x2 = -log(ϵ/2) - log(-log(ϵ/2)) - log(-log(ϵ/2)) / log(ϵ/2) # root of e^-x == x * ϵ/2
     return F.((x0, x1, x2))
 end
-
-#=
-Tighter thresholds can be obtained numerically for common types.
-In particular a tight x2 can be obtained by finding the root of:
-
-log(1 + exp(-x)) / log(1 + exp(x)) == eps(T) / 2
-
-which is the relative error between log(1 + exp(x)) and the approximation x.
-=#
-@inline _log1pexp_thresholds(::Float64) = (-37.0, 18.0, 33.0) # original bounds from Maechler 2012
-@inline _log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 13.9967f0)
-@inline _log1pexp_thresholds(::Float16) = (Float16(-7.625), Float16(3.465), Float16(5.85535))
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -184,7 +184,7 @@ where the tolerances of the approximations ≈ are on the order of eps(T).
     ϵ = big(eps(T))
     x0 = log(ϵ / 2)
     x1 = -log(ϵ) / 2
-    x2 = -x0 - log(-x0) * (1 + 1 / x0) # ≈ root of e^-x == x * ϵ/2
+    x2 = -x0 - log(-x0) * (1 + 1 / x0) # ≈ root of e^-x == x * ϵ/2 via asymptotics of Lambert's W function
     return (T(x0), T(x1), T(x2))
 end
 #= For Float64, Float32 we can hard-code the thresholds to (hopefully) reduce compilation
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -193,9 +193,10 @@ For types for which `precision(x)` depends only on the type of `x`, the compiler
 should optimize away all computations done here.
 =#
 @inline function _log1pexp_thresholds(x::Real)
-    z = (precision(x) - 1) * oftype(x, IrrationalConstants.logtwo)
-    x0 = -z - IrrationalConstants.logtwo
-    x1 = z / 2
+    prec = precision(x)
+    logtwo = oftype(x, IrrationalConstants.logtwo)
+    x0 = -prec * logtwo
+    x1 = (prec - 1) * logtwo / 2
     x2 = -x0 - log(-x0) * (1 + 1 / x0) # approximate root of e^-x == x * ϵ/2 via asymptotics of Lambert's W function
     return (x0, x1, x2)
 end
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -171,14 +171,14 @@ function log1pexp(_x::Real)
 end
 
 #=
-returns a thresholds x0, x1, x2 such that:
+Returns a thresholds x0, x1, x2 such that:
 
     * log1pexp(x) ≈ exp(x) for x ≤ x0
     * log1pexp(x) ≈ log1p(exp(x)) for x0 < x ≤ x1
     * log1pexp(x) ≈ x + exp(-x) for x1 < x ≤ x2
     * log1pexp(x) ≈ x for x > x2
 
-where the tolerances of the approximations ≈ are on the order of eps(T)
+where the tolerances of the approximations ≈ are on the order of eps(T).
 =#
 @inline @generated function _log1pexp_thresholds(::T) where {T<:Real}
     ϵ = big(eps(T))
@@ -187,6 +187,12 @@ where the tolerances of the approximations ≈ are on the order of eps(T)
     x2 = -x0 - log(-x0) * (1 + 1 / x0) # ≈ root of e^-x == x * ϵ/2
     return (T(x0), T(x1), T(x2))
 end
+#= For Float64, Float32 we can hard-code the thresholds to (hopefully) reduce compilation
+times. We will use here the exact same thresholds that were used before the function
+_log1pexp_thresholds (which outputs close but not identical threshods) was introduced,
+to decrease the risk of introducing breaking changes. =#
+@inline _log1pexp_thresholds(::Float64) = (-37e0, 18e0, 33e0)
+@inline _log1pexp_thresholds(::Float32) = (-17f0, 9f0, 16f0)
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -171,10 +171,13 @@ end
 
 #=
 returns a thresholds x0, x1, x2 such that:
+
     * log1pexp(x) ≈ exp(x) for x ≤ x0
     * log1pexp(x) ≈ log1p(exp(x)) for x0 < x ≤ x1
     * log1pexp(x) ≈ x + exp(-x) for x1 < x ≤ x2
     * log1pexp(x) ≈ x for x > x2
+
+where the tolerances of the approximations ≈ are on the order of eps(T)
 =#
 @inline @generated function _log1pexp_thresholds(::T) where {T<:Real}
     F = float(T)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -193,7 +193,7 @@ For types for which `precision(x)` depends only on the type of `x`, the compiler
 should optimize away all computations done here.
 =#
 @inline function _log1pexp_thresholds(x::Real)
-    z = (precision(x) - 1) * oftype(float(x), IrrationalConstants.logtwo)
+    z = (precision(x) - 1) * oftype(x, IrrationalConstants.logtwo)
     x0 = -z - IrrationalConstants.logtwo
     x1 = z / 2
     x2 = -x0 - log(-x0) * (1 + 1 / x0) # approximate root of e^-x == x * ϵ/2 via asymptotics of Lambert's W function
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -159,20 +159,32 @@ See:
 Note: different than Maechler (2012), also uses bounds specific to Float32 and Float16.
 """
 function log1pexp(x::Real)
-    threshold = _log1pexp_threshold(x)
-    if x > threshold # log1pexp(x) ≈ x for large x
-        return oftype(threshold, x)
+    thresh_lin = _log1pexp_threshold_linear(x)
+    if x > thresh_lin
+        return oftype(thresh_lin, x)
+    elseif x > _log1pexp_threshold_exp(x)
+        return x + exp(-x)
     else
     	return log1p(exp(x))
   	end
 end
 
 # returns a threshold such that log1pexp(x) ≈ x for x > threshold
-@generated function _log1pexp_threshold(::T) where {T<:Real}
+@generated function _log1pexp_threshold_linear(::T) where {T<:Real}
     F = float(T)
-    ϵ = big(eps(F)) / 2
-    thresh = -log(expm1(ϵ))
-    return convert(F, thresh)
+    thresh = -log(expm1(big(eps(F)) / 2))
+    return convert(F, max(thresh, log(big(ℯ) - 1)))
+end
+# tighter thresholds can be obtained numerically. Since this is the fastest branch,
+# we do it here for common types.
+@generated _log1pexp_threshold_linear(::Float64) = 33.3e0
+@generated _log1pexp_threshold_linear(::Float32) = 15.9f0
+
+# returns a threshold such that log1pexp(x) ≈ x + exp(-x) for x > threshold
+@generated function _log1pexp_threshold_exp(::T) where {T<:Real}
+    F = float(T)
+    thresh = -log(big(eps(F))) / 2
+    return convert(F, max(thresh, log(big(ℯ) - 1)))
 end
 
 """
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -176,7 +176,7 @@ returns a thresholds x0, x1, x2 such that:
     * log1pexp(x) ≈ x + exp(-x) for x1 < x ≤ x2
     * log1pexp(x) ≈ x for x > x2
 =#
-@generated function _log1pexp_thresholds(::T) where {T<:Real}
+@inline @generated function _log1pexp_thresholds(::T) where {T<:Real}
     F = float(T)
     ϵ = big(eps(F))
     x0 = log(ϵ / 2)
@@ -193,9 +193,9 @@ log(1 + exp(-x)) / log(1 + exp(x)) == eps(T) / 2
 
 which is the relative error between log(1 + exp(x)) and the approximation x.
 =#
-_log1pexp_thresholds(::Float64) = (-37.0, 18.0, 33.0) # original bounds from Maechler 2012
-_log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 13.9967f0)
-_log1pexp_thresholds(::Float16) = (Float16(-7.625), Float16(3.465), Float16(5.85535))
+@inline _log1pexp_thresholds(::Float64) = (-37.0, 18.0, 33.0) # original bounds from Maechler 2012
+@inline _log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 13.9967f0)
+@inline _log1pexp_thresholds(::Float16) = (Float16(-7.625), Float16(3.465), Float16(5.85535))
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -174,37 +174,6 @@ _log1pexp_thresh(::Float16) = Float16(6.24)
 _log1pexp_thresh(::BigFloat) = 172.5936479594263820448907982430859654507995334557035582760493223638550118704546
 _log1pexp_thresh(::Real) = _log1pexp_thresh(0.0)
 
-# _log1p(x::Real) = log1p(x)
-# _log1p(x::Real, thresh::Real) = abs(x) ≤ thresh ? x : oftype(x, log1p(x))
-# _log1p(x::Float16) = _log1p(x, exp(-7))
-# _log1p(x::Float32) = _log1p(x, exp(-15))
-# _log1p(x::Float64) = _log1p(x, exp(-37))
-
-# function log1pexp(x::Union{Float16,Float32,Float64})
-# 	a, b, c = _log1pexp_branch_bounds(x)
-#     if x ≤ a
-#         return exp(x)
-#     elseif x ≤ b
-#         return log1p(exp(x))
-#     elseif x ≤ c
-#         return x + exp(-x)
-#     else
-#         return x
-#     end
-# end
-
-#=
-Given the `approx` used in a branch of log1pexp(x) above, we find the first `x` (from above
-or below) that is a root of
-
-    T(log1pexp(big(x))) - approx(T(x))
-
-This determines the branch bounds below.
-=#
-@inline _log1pexp_branch_bounds(::Float64) = (-37.0, 18.0, 33.3)
-@inline _log1pexp_branch_bounds(::Float32) = (-15f0, 9f0, 14.5f0)
-@inline _log1pexp_branch_bounds(::Float16) = Float16.((-7, 3, 5.7))
-
 """
 $(SIGNATURES)
 
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -172,13 +172,7 @@ end
 
 #= The precision of BigFloat cannot be computed from the type only and computing
 thresholds is slow. Therefore prefer version without thresholds in this case. =#
-function log1pexp(x::BigFloat)
-    if x > 0
-        return x + log1p(exp(-x))
-    else
-        return log1p(exp(x))
-    end
-end
+log1pexp(x::BigFloat) = x > 0 ? x + log1p(exp(-x)) : log1p(exp(x))
 
 #=
 Returns thresholds x0, x1, x2 such that:
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -156,8 +156,11 @@ transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`log
 See:
  * Martin Maechler (2012) [“Accurately Computing log(1 − exp(− |a|))”](http://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf)
 """
-function log1pexp(_x::Real)
-    x = float(_x)
+log1pexp(x::Real) = _log1pexp(float(x)) # ensures that BigInt/BigFloat, Int/Float64 etc. dispatch to the same algorithm
+
+# Approximations based on Maechler (2012)
+# Argument `x` is a floating point number due to the definition of `log1pexp` above
+function _log1pexp(x::Real)
     x0, x1, x2 = _log1pexp_thresholds(x)
     if x < x0
         return exp(x)
@@ -172,7 +175,7 @@ end
 
 #= The precision of BigFloat cannot be computed from the type only and computing
 thresholds is slow. Therefore prefer version without thresholds in this case. =#
-log1pexp(x::BigFloat) = x > 0 ? x + log1p(exp(-x)) : log1p(exp(x))
+_log1pexp(x::BigFloat) = x > 0 ? x + log1p(exp(-x)) : log1p(exp(x))
 
 #=
 Returns thresholds x0, x1, x2 such that:
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -159,20 +159,16 @@ See:
 Note: different than Maechler (2012), also uses bounds specific to Float32 and Float16.
 """
 function log1pexp(x::Real)
-    if x > _log1pexp_thresh(x)
-        return float(x)
+    threshold = _log1pexp_threshold(x)
+    if x > threshold # log1pexp(x) ≈ x for large x
+        return oftype(threshold, x)
     else
-        t = log1p(exp(-abs(x)))
-        return x ≤ 0 ? t : t + x
-    end
+    	return log1p(exp(x))
+  	end
 end
 
-# threshold `thresh` such that `log1pexp(x) == x` for `x > thresh`
-_log1pexp_thresh(::Float64) = 33.27106466687738
-_log1pexp_thresh(::Float32) = 14.556091f0
-_log1pexp_thresh(::Float16) = Float16(6.24)
-_log1pexp_thresh(::BigFloat) = 172.5936479594263820448907982430859654507995334557035582760493223638550118704546
-_log1pexp_thresh(::Real) = _log1pexp_thresh(0.0)
+# returns a threshold such that log1pexp(x) ≈ x for x > threshold
+@generated _log1pexp_threshold(::T) where {T<:Real} = -log(expm1(eps(float(T))/2))
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -171,7 +171,7 @@ function log1pexp(_x::Real)
 end
 
 #=
-Returns a thresholds x0, x1, x2 such that:
+Returns thresholds x0, x1, x2 such that:
 
     * log1pexp(x) ≈ exp(x) for x ≤ x0
     * log1pexp(x) ≈ log1p(exp(x)) for x0 < x ≤ x1
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -152,9 +152,58 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
 
 This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref).
+
+See:
+ * Martin Maechler (2012) [“Accurately Computing log(1 − exp(− |a|))”](http://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf)
+
+Note: different than Maechler (2012), also uses bounds specific to Float32 and Float16.
 """
-log1pexp(x::Real) = x ≤ -37 ? exp(x) : x ≤ 18 ? log1p(exp(x)) : x ≤ 33.3 ? x + exp(-x) : float(x)
-log1pexp(x::Float32) = x < 9f0 ? log1p(exp(x)) : x < 16f0 ? x + exp(-x) : oftype(exp(-x), x)
+function log1pexp(x::Real)
+    if x > _log1pexp_thresh(x)
+        return float(x)
+    else
+        t = log1p(exp(-abs(x)))
+        return x ≤ 0 ? t : t + x
+    end
+end
+
+# threshold `thresh` such that `log1pexp(x) == x` for `x > thresh`
+_log1pexp_thresh(::Float64) = 33.27106466687738
+_log1pexp_thresh(::Float32) = 14.556091f0
+_log1pexp_thresh(::Float16) = Float16(6.24)
+_log1pexp_thresh(::BigFloat) = 172.5936479594263820448907982430859654507995334557035582760493223638550118704546
+_log1pexp_thresh(::Real) = _log1pexp_thresh(0.0)
+
+# _log1p(x::Real) = log1p(x)
+# _log1p(x::Real, thresh::Real) = abs(x) ≤ thresh ? x : oftype(x, log1p(x))
+# _log1p(x::Float16) = _log1p(x, exp(-7))
+# _log1p(x::Float32) = _log1p(x, exp(-15))
+# _log1p(x::Float64) = _log1p(x, exp(-37))
+
+# function log1pexp(x::Union{Float16,Float32,Float64})
+# 	a, b, c = _log1pexp_branch_bounds(x)
+#     if x ≤ a
+#         return exp(x)
+#     elseif x ≤ b
+#         return log1p(exp(x))
+#     elseif x ≤ c
+#         return x + exp(-x)
+#     else
+#         return x
+#     end
+# end
+
+#=
+Given the `approx` used in a branch of log1pexp(x) above, we find the first `x` (from above
+or below) that is a root of
+
+    T(log1pexp(big(x))) - approx(T(x))
+
+This determines the branch bounds below.
+=#
+@inline _log1pexp_branch_bounds(::Float64) = (-37.0, 18.0, 33.3)
+@inline _log1pexp_branch_bounds(::Float32) = (-15f0, 9f0, 14.5f0)
+@inline _log1pexp_branch_bounds(::Float16) = Float16.((-7, 3, 5.7))
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -185,9 +185,17 @@ returns a thresholds x0, x1, x2 such that:
     return F.((x0, x1, x2))
 end
 
-# tighter thresholds can be obtained numerically for common types
-_log1pexp_thresholds(::Float64) = (-37.0, 18.0, 33.0)
-_log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 15.9458f0)
+#=
+Tighter thresholds can be obtained numerically for common types.
+In particular a tight x2 can be obtained by finding the root of:
+
+log(1 + exp(-x)) / log(1 + exp(x)) == eps(T) / 2
+
+which is the relative error between log(1 + exp(x)) and the approximation x.
+=#
+_log1pexp_thresholds(::Float64) = (-37.0, 18.0, 33.0) # original bounds from Maechler 2012
+_log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 13.9967f0)
+_log1pexp_thresholds(::Float16) = (Float16(-7.625), Float16(3.465), Float16(5.85535))
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -195,14 +195,16 @@ should optimize away all computations done here.
     return (x0, x1, x2)
 end
 
-#= For Float64, Float32 we can hard-code the thresholds to make absolutely sure they are not
+#=
+For common types we hard-code the thresholds to make absolutely sure they are not
 recomputed each time. Also, _log1pexp_thresholds is not completely elided by the
-compiler in Julia 1.0 / 1.6 which LogExpFunctions intends to support.
-For Float64 we use the same exact thresholds given by Maechler 2012,
-since these were the ones used before _log1pexp_thresholds was introduced (which outputs
-close but not identical thresholds), to reduce any (small) risk of breakage.
-For Float32 and Float16 we use truncated versions of the output of _log1pexp_thresholds. =#
-@inline _log1pexp_thresholds(::Float64) = (-37e0, 18e0, 33e0)
+compiler in Julia 1.0 / 1.6. We round the outputs of the non-hardcoded version:
+
+* _log1pexp_thresholds(::Float64) = (-36.7368005696771, 18.021826694558577, 33.23111882352963)
+* _log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 13.993f0)
+* _log1pexp_thresholds(::Float16) = (-7.625, 3.467, 5.86)
+=#
+@inline _log1pexp_thresholds(::Float64) = (-37e0, 18e0, 33e0) # same as Maechler 2012
 @inline _log1pexp_thresholds(::Float32) = (-17f0, 8f0, 14f0)
 @inline _log1pexp_thresholds(::Float16) = (Float16(-7.6), Float16(3.5), Float16(5.9))
 
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -189,7 +189,7 @@ where the tolerances of the approximations ≈ are on the order of eps(T).
 end
 #= For Float64, Float32 we can hard-code the thresholds to (hopefully) reduce compilation
 times. We will use here the exact same thresholds that were used before the function
-_log1pexp_thresholds (which outputs close but not identical threshods) was introduced,
+_log1pexp_thresholds (which outputs close but not identical thresholds) was introduced,
 to decrease the risk of introducing breaking changes. =#
 @inline _log1pexp_thresholds(::Float64) = (-37e0, 18e0, 33e0)
 @inline _log1pexp_thresholds(::Float32) = (-17f0, 9f0, 16f0)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -196,17 +196,12 @@ should optimize away all computations done here.
 end
 
 #=
-For common types we hard-code the thresholds to make absolutely sure they are not
-recomputed each time. Also, _log1pexp_thresholds is not completely elided by the
-compiler in Julia 1.0 / 1.6. We round the outputs of the non-hardcoded version:
-
-* _log1pexp_thresholds(::Float64) = (-36.7368005696771, 18.021826694558577, 33.23111882352963)
-* _log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 13.993f0)
-* _log1pexp_thresholds(::Float16) = (-7.625, 3.467, 5.86)
+For common types we hard-code the thresholds to make absolutely sure they are not recomputed
+each time. Also, _log1pexp_thresholds is not elided by the compiler in Julia 1.0 / 1.6.
 =#
-@inline _log1pexp_thresholds(::Float64) = (-37e0, 18e0, 33e0) # same as Maechler 2012
-@inline _log1pexp_thresholds(::Float32) = (-17f0, 8f0, 14f0)
-@inline _log1pexp_thresholds(::Float16) = (Float16(-7.6), Float16(3.5), Float16(5.9))
+@inline _log1pexp_thresholds(::Float64) = (-36.7368005696771, 18.021826694558577, 33.23111882352963)
+@inline _log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 13.993f0)
+@inline _log1pexp_thresholds(::Float16) = (Float16(-7.625), Float16(3.467), Float16(5.86))
 
 """
 $(SIGNATURES)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -170,6 +170,17 @@ function log1pexp(_x::Real)
     end
 end
 
+#= The precision of BigFloat cannot be computed from the type only.
+Since computing the dynamically thresholds is slow, prefer to use
+a slower but accurate implementation in this case. =#
+function log1pexp(x::BigFloat)
+    if x > 0
+        return x + log1p(exp(-x))
+    else
+        return log1p(exp(x))
+    end
+end
+
 #=
 Returns thresholds x0, x1, x2 such that:
 
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -178,19 +178,23 @@ Returns thresholds x0, x1, x2 such that:
     * log1pexp(x) ≈ x + exp(-x) for x1 < x ≤ x2
     * log1pexp(x) ≈ x for x > x2
 
-where the tolerances of the approximations ≈ are on the order of eps(T).
+where the tolerances of the approximations are on the order of eps(typeof(x)).
+For types for which `precision(x)` depends only on the type of `x`, the compiler
+should optimize away all computations done here.
 =#
-@inline @generated function _log1pexp_thresholds(::T) where {T<:Real}
-    ϵ = big(eps(T))
-    x0 = log(ϵ / 2)
-    x1 = -log(ϵ) / 2
-    x2 = -x0 - log(-x0) * (1 + 1 / x0) # ≈ root of e^-x == x * ϵ/2 via asymptotics of Lambert's W function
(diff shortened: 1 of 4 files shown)


cossio:
Benchmarks.

```julia
using BenchmarkTools, LogExpFunctions
log1pexp_old(x::Float64) = x < 18 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x) # before this PR
log1pexp_new(x::Float64) = x ≤ -37 ? exp(x) : x ≤ 18 ? log1p(exp(x)) : x ≤ 33.3 ? x + exp(-x) : float(x) # this PR


julia> @benchmark log1pexp_old(x) setup=(x=(rand() - 0.5) * 100) samples=10^6                                                                                                                                                             
BenchmarkTools.Trial: 240558 samples with 998 evaluations.                  
 Range (min … max):   9.675 ns … 136.391 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     19.892 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   19.025 ns ±   7.230 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▁█                       ▂▄                   ▅             
  ▃███▆▃▁▁▃▁▃▄█▂▂▁▁▁▁▁▁▁▅▃▆▆▅██▁▃▁▁▁▁▁▁▁▂▁▃▁▄▂▁▆▁▁█▁▁▂▁▁▁▁▁▁▁▁ ▂
  9.68 ns         Histogram: frequency by time         35.2 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.            


julia> @benchmark log1pexp_new(x) setup=(x=(rand() - 0.5) * 100) samples=10^6                                                                                                                                                             
BenchmarkTools.Trial: 281295 samples with 1000 evaluations.                 
 Range (min … max):   2.794 ns … 343.354 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     18.832 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.156 ns ±   8.600 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                               ▄ ▄                           
  █▂▁▁▁▁▁▁▁▁▁▁▁▆▇▂▇▆▃▂▂▁▁▁▁▁▁▁▁▁▁▁█▂█▆▁▃▁▂▁▁▁▁▁▁▁▁█▁█▁▄▁▂▁▁▂▁▁ ▂
  2.79 ns         Histogram: frequency by time         32.1 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

test/basicfuns.jl
```diff
@@ -110,15 +110,17 @@ end
...
-    @test log1pexp(-10000) ≈ 0.0
-
-    @test log1pexp(2f0)      ≈ log(1f0 + exp(2f0))
-    @test log1pexp(-2f0)     ≈ log(1f0 + exp(-2f0))
-    @test log1pexp(10000f0)  ≈ 10000f0
-    @test log1pexp(-10000f0) ≈ 0f0
+    # test every branch
+    for x in (0, 1, 2, 10, 20, 40), T in (Int, Float32, Float64)
+        @test (@inferred log1pexp(-T(x))) ≈ log1p(exp(big(-x)))
+        @test (@inferred log1pexp(+T(x))) ≈ log1p(exp(big(+x)))
```
tpapp, line 116: thanks for reorganizing the tests, too

src/basicfuns.jl
```diff
@@ -153,8 +153,8 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
 This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref).
 """
-log1pexp(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)
-log1pexp(x::Float32) = x < 9.0f0 ? log1p(exp(x)) : x < 16.0f0 ? x + exp(-x) : oftype(exp(-x), x)
+log1pexp(x::Real) = x ≤ -37 ? exp(x) : x ≤ 18 ? log1p(exp(x)) : x ≤ 33.3 ? x + exp(-x) : float(x)
```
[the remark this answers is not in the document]
cossio: The `oftype(exp(-x), x)` version seems to be much slower, see benchmark in my comment below.

cossio:
Benchmark comparing to version with `oftype(...)` instead of `float(...)` suggested by @tpapp .

```julia
log1pexp_oftype(x::Float64) = x ≤ -37 ? exp(x) : x ≤ 18 ? log1p(exp(x)) : x ≤ 33.3 ? x + exp(-x) : oftype(exp(x), x)


julia> @benchmark log1pexp_new(x) setup=(x=(rand() - 0.5) * 100) samples=10^6                                                                                                                                                             
BenchmarkTools.Trial: 272085 samples with 1000 evaluations.
 Range (min … max):   2.460 ns …  1.768 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     19.383 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.689 ns ± 11.719 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                ▁                 ▃▂                        
  █▄▂▂▂▂▂▂▂▂▂▂▁▂▇▅▄█▇▅▄▂▂▂▂▂▂▂▂▂▂▂▂▆▃██▂▇▃▆▂▂▂▂▂▂▂▂▅▂█▂▇▂▅▃▂▄ ▃
  2.46 ns         Histogram: frequency by time        31.2 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark log1pexp_oftype(x) setup=(x=(rand() - 0.5) * 100) samples=10^6                                                                                                                                                          
BenchmarkTools.Trial: 233832 samples with 998 evaluations.
 Range (min … max):   9.697 ns … 833.891 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     20.884 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   19.594 ns ±   8.968 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▅█▇▆               ▆ █ ▄          ▁ ▂                      
  ▂▅▇█████▃▁▁▁▁▁▁▁▁▁▁▂▄▆█▂█▅█▂▃▁▁▁▁▁▂▅▁█▁█▁▇▄▁▃▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  9.7 ns          Histogram: frequency by time         39.7 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

File: src/basicfuns.jl (hunk 2 above)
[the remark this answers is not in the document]
Comment (tpapp) [reply src/basicfuns.jl:156]: OK, then `float` should be fine. With the assumption that if someone runs into an inference problem with a particular user-defined type, they should first make sure that `typeof(float(x)) == typeof(exp(x))`, which is reasonable to expect anyway and should hold for all base types.

**Review by tpapp — APPROVED** [review]
LGTM.

File: src/basicfuns.jl (hunk 2 above)
Comment (devmotion) [inline src/basicfuns.jl:156]: Can we use `<=` or `<` consistently kn both the Float32 and Float64 version?

**Review by devmotion — CHANGES_REQUESTED** [review]
Can you update the ChainRules tests as well? IIRC there we also test every branch.

My main concern is whether there is a licensing issue since this is based on (the vignette of) Rmpfr which uses a GPL. I would prefer if someone who's more familiar with licenses (I'm definitely not) could confirm that there are no problems here and we don't have use the GPL.

File: src/basicfuns.jl (hunk 2 above)
Comment (cossio) [reply src/basicfuns.jl:156]: I went with `<=` in the Float32 version.

Comment (devmotion) [reply src/basicfuns.jl:156]: 👍 It would also be good to add a link to the reference the implementation is based on. I think for some othef functions it's included in the docstring, so we could do the same here.

**Comment by cossio** [comment]
The ChainRules tests fail for `Float32`. But I noticed that `logistic` itself has no tests for `Float32`. Are we sure its implementation is correct?

**Comment by cossio** [comment]
I unified the `log1pexp` implementations for `Float32` and `Float64`. Here is how I computed the branch bounds for the different approximations:

```julia
function log1pexp(x::Real)
    t = log1p(exp(-abs(x)))
    return x ≤ 0 ? t : t + x
end

xs = 0:-0.01:-100
for T in (Float16, Float32, Float64)
	for x in xs
		correct = T(log1pexp(big(x)))
		if iszero(correct - exp(T(x)))
		   println("Found crossing with `exp(x)` at x = $x for $T, in the interval $(extrema(xs))")
		   break
		end
	end
end

xs = 0:0.01:100
for T in (Float16, Float32, Float64)
	for x in xs
		correct = T(log1pexp(big(x)))
		if iszero(correct - (T(x) + exp(-T(x))))
		   println("Found crossing with `x + exp(-x)` at x = $x for $T, in the interval $(extrema(xs))")
		   break
		end
	end
end

xs = 0:0.01:100
for T in (Float16, Float32, Float64)
	for x in xs
		if iszero(T(log1pexp(big(x))) - T(x))
		   println("Found crossing with `x` at x = $x for $T, in the interval $(extrema(xs))")
		   break
		end
	end
end
```

I removed the ChainRules tests for `Float32`, since the chain rule itself is not specialized (it relies on `logistic`).

src/basicfuns.jl
```diff
@@ -153,8 +153,37 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
 This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref).
 """
-log1pexp(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)
-log1pexp(x::Float32) = x < 9.0f0 ? log1p(exp(x)) : x < 16.0f0 ? x + exp(-x) : oftype(exp(-x), x)
+function log1pexp(x::Real)
+    t = log1p(exp(-abs(x)))
+    return x ≤ 0 ? t : t + x
+end
```
devmotion, line 159: What's the reason for this change? I am worried this will reduce accuracy eg for `ForwardDiff.Dual`.

src/basicfuns.jl
```diff
@@ -153,8 +153,37 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+
+function log1pexp(x::Union{Float16,Float32,Float64})
+    #= Faster implementation for floats. Based on Mächler (2012), available at:
+    https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf =#
```
devmotion, line 163: Can you move this reference to the docstring instead, similar to `log1mexp`?

src/basicfuns.jl
```diff
@@ -153,8 +153,37 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+Given the `approx` used in a branch of log1pexp(x) above, we find the first `x` (from above
+or below) that is a root of
+
+    T(log1pexp(big(x))) - approx(T(x))
+
+This determines the branch bounds below.
+=#
+_log1pexp_branch_bounds(::Float64) = (-37, 18, 33.3)
+_log1pexp_branch_bounds(::Float32) = (-15, 9, 14.5)
+_log1pexp_branch_bounds(::Float16) = (-7, 3, 5.7)
```
devmotion, line 186: Are these the same as reported in the reference? It is probably better to make them all of type Float64, Float32, and Float16 to avoid promotions in the conditional statements.

test/chainrules.jl
```diff
@@ -57,14 +57,10 @@
...
 
-    # test all branches of `log1pexp`
-    for x in (-20.9, 15.4, 41.5)
-        test_frule(log1pexp, x)
-        test_rrule(log1pexp, x)
-    end
-    for x in (8.3f0, 12.5f0, 21.2f0)
-        test_frule(log1pexp, x; rtol=1f-3, atol=1f-3)
-        test_rrule(log1pexp, x; rtol=1f-3, atol=1f-3)
+    @testset "log1pexp on $T" for T in (Float64, BigFloat)
```
devmotion, line 60: Why did you not check Float32 as well? The main "issue" there should be that one has to adjust the tolerances, as in the current tests, due to issues/inaccuracies of FiniteDifferences with Float32.

devmotion (changes requested):

src/basicfuns.jl
cossio: The bounds given before are specific to Float64, and should be inaccurate for other types. I prefer a more robust method as fallback. Note that this implementation is very accurate as far as I can tell, it's just slower.

devmotion: The point is that this fallback will be hit now whereas before a, for 64bit floating point numbers, more optimized version was used. And this will affect eg. `ForwardDiff.Dual`. The problem is not the accuracy of the Float64/32/16 implementations but the reduced accuracy for other types.

test/chainrules.jl
cossio: Ok I adjusted the tols and now I am checking Float32 too. Thanks!

src/basicfuns.jl
cossio: I would have thought the compiler would take care of the promotions (isn't this some form of constant propagation?)

src/basicfuns.jl
cossio: Yep

src/basicfuns.jl
devmotion: My suggestion is to not touch the default definition in this PR since it requires caeeful evaluation for which types accuracy and performance would be reduced, and for which maybe it would be improved.

cossio: I'm not sure I follow. `log1pexp(x::Real)` is the most accurate version, it is just slow.

So there can be an issue that `log1pexp(::ForwardDiff.Dual)` will now be slower. Is this your concern?

If that's so I agree. But then the point is that the fast implementation is using bounds tailored for Float64. So we can't just have that method be the fallback for all `Real`s, right?

src/basicfuns.jl
```diff
@@ -152,9 +152,39 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf.
 """
-log1pexp(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)
-log1pexp(x::Float32) = x < 9.0f0 ? log1p(exp(x)) : x < 16.0f0 ? x + exp(-x) : oftype(exp(-x), x)
+function log1pexp(x::Real)
+    t = log1p(exp(-abs(x)))
+    return x ≤ 0 ? t : t + x
+end
+
+function log1pexp(x::Union{Float16,Float32,Float64})
```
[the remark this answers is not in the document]
cossio: What do you mean? Can you give an example on how extendability is limited by this approach?

src/basicfuns.jl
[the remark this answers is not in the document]
cossio: Please see the current version.

src/basicfuns.jl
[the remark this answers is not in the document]
devmotion: You define a function for `::Union{Float64, Float32, Float16}` - that's totally fine but it means you can't reuse the logic and just define new bounds for other types and it works. So the possibility to extend the function for the bounds is not really useful in practice, and instead one could just hardcode them in the function body. But to save redundant code one could use `@eval`, eg.
```julia
for (T, a, b, c) in ((:Float64, -37.0, ...), ...)
    @eval begin
         function log1pexp(x::$T)
              if x < $a
...
```

test/chainrules.jl
```diff
@@ -57,14 +57,11 @@
...
-        test_frule(log1pexp, x)
-        test_rrule(log1pexp, x)
-    end
-    for x in (8.3f0, 12.5f0, 21.2f0)
-        test_frule(log1pexp, x; rtol=1f-3, atol=1f-3)
-        test_rrule(log1pexp, x; rtol=1f-3, atol=1f-3)
+    @testset "log1pexp" begin
+        for x in (0, 1, 2, 10, 15, 20, 40), s in (-1, 1)
+            test_scalar(log1pexp, Float64(s * x))
+            test_scalar(log1pexp, Float32(s * x); rtol=1f-3, atol=1f-3)
```
devmotion, line 63:
```suggestion
        for absx in (0, 1, 2, 10, 15, 20, 40), x in (-absx, absx)
            test_scalar(log1pexp, Float64(x))
            test_scalar(log1pexp, Float32(x); rtol=1f-3, atol=1f-3)
```

src/basicfuns.jl
devmotion: We want to move to this fallback definition, I think - but there are cases where the current fallback is accurate and fast, even though inputs are not Float64. And I don't think it's acceptable to introduce performance regressions in these cases, and in particular not in a PR where it is not necessary to change the fallback definition.

I suggest keeping it for now, opening an issue, evaluating which types are affected by the change, and then e.g. overloading `log1pexp` in ForwardDiff if needed, and finally replacing the fallback.

test/chainrules.jl
cossio: Nice.

src/basicfuns.jl
```diff
@@ -152,9 +152,43 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+Given the `approx` used in a branch of log1pexp(x) above, we find the first `x` (from above
+or below) that is a root of
+
+    T(log1pexp(big(x))) - approx(T(x))
+
+This determines the branch bounds below.
+=#
+_log1pexp_branch_bounds(::Float64) = (-37.0, 18.0, 33.3)
+_log1pexp_branch_bounds(::Float32) = (-15f0, 9f0, 14.5f0)
+_log1pexp_branch_bounds(::Float16) = (Float16(-7), Float16(3), Float16(5.7))
```
devmotion, line 191: This seems a bit complicated, in particular I don't think we should introduce so many levels of functions. Maybe just use

```suggestion
# The thresholds below are determined by the first root `x` (from above or below)
#of `T(log1pexp(big(x))) - approx(T(x))``, where `approx` is the approximation of interest.
for (T, a, b, c) in ((:Float64, -37.0, 18.0, 33.3), (:Float32, -15f0, 9f0, 14.5f0), (:Float16, Float16(-7), Float16(3), Float16(5.7))
    @eval function function log1pexp(x::$T)
        if x ≤ $a
            return exp(x)
        elseif x ≤ $b
            return log1p(exp(x))
        elseif x ≤ $c
            return x + exp(-x)
        else
            return x
        end
    end
end
```

cossio:
As to the performance issue, let's compare the current implementation on master with the generic `log1pexp(x::Real)` of this PR:

```julia
using BenchmarkTools

# generic fallback in this PR
function log1pexp_new(x::Real)
    t = log1p(exp(-abs(x)))
    return x ≤ 0 ? t : t + x
end

# current master
log1pexp_old(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)


julia> @benchmark log1pexp_old(x) setup=(x=(rand() - 0.5) * 100) samples=10^6                                                                                                                                                             
BenchmarkTools.Trial: 278000 samples with 999 evaluations.     
 Range (min … max):   9.042 ns … 56.144 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     17.880 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.501 ns ±  6.132 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▆█▇▆▄▃▃ ▂▁▇▆▄▃▃▁▂▃▂▁    ▃▆█▅▇▄▅▁▃ ▂▂▁▃  ▁   ▃ ▅▇ ▆▂ ▅▁ ▃  ▂ ▄
  ████████████████████▇▇▆▇███████████████▇█▆▇▇█▆██▇██▇██▇██▆█ █
  9.04 ns      Histogram: log(frequency) by time      29.1 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.                                                                                                                                                                                            


julia> @benchmark log1pexp_new(x) setup=(x=(rand() - 0.5) * 100) samples=10^6                                                                                                                                                             
BenchmarkTools.Trial: 218739 samples with 997 evaluations.
 Range (min … max):  12.472 ns … 1.042 μs  ┊ GC (min … max): 0.00% … 0.00%                                                                                               
 Time  (median):     21.477 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   21.094 ns ± 8.230 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                          ▇ █                                 
  ▁▃▄▄▂▅▇▁▂▁▁▁▁▁▁▁▂▁▇▂▇▄▇▂█▁█▂▂▂▁▂▁▁▂▂▁▁▂▁▁▂▁▂▂▁▁▂▁▁▁▁▁▁▁▁▁▁ ▂                                                                                                                                                                            
  12.5 ns        Histogram: frequency by time          36 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

src/basicfuns.jl
cossio: Hmm I'd prefer to go back to

```
function log1pexp(x::Union{Float16,Float32,Float64})
	a, b, c = _log1pexp_branch_bounds(x)
    if x ≤ a
        return exp(x)
    elseif x ≤ b
        return log1p(exp(x))
    elseif x ≤ c
        return x + exp(-x)
    else
        return x
    end
end
```

It's the same pattern used by `logistic`.

devmotion:
No, it doesn't convince me. The `Float16` issue is already fixed by the PR. IMO there's no reason to change the fallback in this PR, it's a separate issue that has to be done carefully to avoid performance regressions: This PR is about adding optimizations/fixes for Float64, Float32, and Float16. LogExpFunctions is a central package in the Julia ecosystem, and hence we should ensure at least that for common floating point number types other than `Float64` in downstream packages there are no sudden surprising regressions. This deserves a separate issue, PR, and possibly downstream changes, and should not be part of this PR.

src/basicfuns.jl
cossio: I still don't understand how what you propose with `@eval` is more "extensible" than this ...

devmotion: No, of course, it's not more extendable. The point is that it avoids defining any other function besides `log1pexp` since these can't be extended either in the version you suggest.

cossio:
I simplified the implementation once more.

Updated benchmarks:

```
julia> @benchmark log1pexp(x) setup=(x=(rand() - 0.5) * 100) samples=10^6        # after this PR                                                                                                                                                          
BenchmarkTools.Trial: 261384 samples with 1000 evaluations.
 Range (min … max):   2.439 ns … 57.839 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     20.283 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   17.496 ns ±  7.741 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▇                                ▃ ▅█ ▄                      
  █▃▂▂▂▂▂▂▂▂▂▂▁▁▁▁▂▄▄▅▄▃▃▂▂▂▂▂▂▂▂▂▂█▂██▂█▅▆▄▂▂▂▂▂▃▃▂▃▂▂▃▂▃▂▂▂ ▃
  2.44 ns         Histogram: frequency by time        32.3 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

whereas the current master takes about 9 secs minimum (see plot above).

src/basicfuns.jl
```diff
@@ -152,9 +152,27 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+        t = log1p(exp(-abs(x)))
+        return x ≤ 0 ? t : t + x
+    end
+end
+
+# threshold `thresh` such that `log1pexp(x) == x` for `x > thresh`
+_log1pexp_thresh(::Float64) = 33.27106466687738
+_log1pexp_thresh(::Float32) = 14.556091f0
+_log1pexp_thresh(::Float16) = Float16(6.24)
+_log1pexp_thresh(::BigFloat) = 172.5936479594263820448907982430859654507995334557035582760493223638550118704546
```
devmotion, line 174: I don't think this is correct in general? Shouldn't this depend on the precision of the BigFloat input?

devmotion (changes requested):

src/basicfuns.jl
```diff
@@ -152,9 +152,27 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
 transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref).
+
+See:
+ * Martin Maechler (2012) [“Accurately Computing log(1 − exp(− |a|))”](http://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf)
+
+Note: different than Maechler (2012), also uses bounds specific to Float32 and Float16.
 """
-log1pexp(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)
-log1pexp(x::Float32) = x < 9.0f0 ? log1p(exp(x)) : x < 16.0f0 ? x + exp(-x) : oftype(exp(-x), x)
+function log1pexp(x::Real)
```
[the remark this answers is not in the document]
cossio: Sorry I don't want to make too much noise. I'm just trying to find an implementation that really works for other floats besides Float64. The current `log1pexp(x::Real)` implementation in master is only accurate for Float64 (the bounds derived in the paper are only given for double precision).

src/basicfuns.jl
```diff
@@ -152,9 +152,27 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+    else
+        t = log1p(exp(-abs(x)))
+        return x ≤ 0 ? t : t + x
+    end
+end
+
+# threshold `thresh` such that `log1pexp(x) == x` for `x > thresh`
+_log1pexp_thresh(::Float64) = 33.27106466687738
+_log1pexp_thresh(::Float32) = 14.556091f0
+_log1pexp_thresh(::Float16) = Float16(6.24)
```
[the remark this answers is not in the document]
cossio: In the experiments I've done it seems that only the threshold `x > thresh` which then branches to just `return x` is relevant for the performance. The other branches are not helping much in terms of speed, so one can just go on to call the most accurate generic implementation for `x < thresh`.

In fact the approximations going on in the original reference for `x < thresh` are just approximating `log1p`. But it seems we have a very optimized `log1p` so maybe these are not necessary.

src/basicfuns.jl
cossio: I'm not sure. Let me think a little more about this I think there can be a generic way to compute this threshold instead of hard-coding it.

src/basicfuns.jl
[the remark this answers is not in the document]
devmotion: I know I said this before - but this PR should really be limited to a small fixed subset of concrete types and the default implementation should not be changed. There are too many things that can go wrong, so it is much cleaner and simpler if we start with Float64/32/16 here.

src/basicfuns.jl
```diff
@@ -152,9 +152,28 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+    threshold = _log1pexp_threshold(x)
+    if x > threshold # log1pexp(x) ≈ x for large x
+        return oftype(threshold, x)
+    else
+    	return log1p(exp(x))
+  	end
+end
+
+# returns a threshold such that log1pexp(x) ≈ x for x > threshold
+@generated function _log1pexp_threshold(::T) where {T<:Real}
```
devmotion, line 171: LogExpFunctions is supposed to be a very very lightweight package. Hence I am against adding generated functions if they can be avoided. Again, there are no such issues if you limit this PR to Float64/32/16.

cossio:
To expand on my previous comments. Here is the approximation algorithm of Maechler (2012):

image

Those branch bounds are expected to be accurate for `Float64`. This is benchmarked in my comment above in https://github.com/JuliaStats/LogExpFunctions.jl/pull/37#issuecomment-1060647460 (called `log1pexp_new` there).

Now, the first branch is just approximating `log1p(y)` with `y` when `y` is small (in absolute value). If we trust our `log1p` is already well optimized, it should take of this by itself. The third branch is doing something similar, `log1pexp(x) == x + log1p(exp(-x)) ≈ x + exp(-x)`, but we can exploit that `x` is being summed to be more aggressive in approximating `log1p(y) ≈ y` than what `log1p` could do by itself, so we leave this branch there. Therefore we simplify to:

image

Next, we can compute generic bounds that apply to other floats besides `Float64`. 

image

image

src/basicfuns.jl
cossio: Mmm I guess something equivalent can be writen with `@eval` instead of `@generated`.

devmotion:
> Now, the first branch is just approximating `log1p(y)` with `y` when `y` is small (in absolute value). If we trust our `log1p` is already well optimized, it should take of this by itself. The third branch is doing something similar, `log1pexp(x) == x + log1p(exp(-x)) ≈ x + exp(-x)`, but we can exploit that `x` is being summed to be more aggressive in approximating `log1p(y) ≈ y` than what `log1p` could do by itself, so we leave this branch there.

It's not about accuracy only, and it's not about speed only. It's about speed AND accuracy and how to trade off one versus the other. So even if using `log1p` is accurate or more accurate than using these two branches, it might be worth using the branches if increases speed without sacrificing accuracy too much.

cossio:
> It's not about accuracy only, and it's not about speed only. It's about speed AND accuracy and how to trade off one versus the other. So even if using `log1p` is accurate or more accurate than using these two branches, it might be worth using the branches if it increases speed without sacrificing accuracy too much.

It doesn't increase speed. Please see benchmarks.

devmotion:
> Please see benchmarks.

I'm sorry, it's really difficult to navigate your comments and benchmarks. This is why I think this should be discussed, and evaluated and summarized separately first. I think it's quite confusing to have these lengthy discussions and explorations in a PR - which initially was focused on something else.

devmotion:
I checked the benchmarks (not sure if the correct ones) but I don't think they allow to draw conclusions about the speed of the branches. They use random numbers which e.g. are < -37 only with around 13% probability. So the results will be biased by the other branches.

cossio:
> I checked the benchmarks (not sure if the correct ones) but I don't think they allow to draw conclusions about the speed of the branches. They use random numbers which e.g. are < -37 only with around 13% probability. So the results will be biased by the other branches.

Good point! Okay I benchmarked only `x < -37` and there is a gain of 2ns by including the benchmark (11ns without the branch vs. 9ns with this branch). Isn't this saying that `log1p` can be optimized in this range?

Anyway, you are right this discussion is too long. I'll follow your advice and just do the `x < -37` here and then I'l make another PR trying to make it generic!

src/basicfuns.jl
```diff
@@ -152,9 +152,50 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+end
+
+#=
+returns a thresholds x0, x1, x2 such that:
+    * log1pexp(x) ≈ exp(x) for x ≤ x0
+    * log1pexp(x) ≈ log1p(exp(x)) for x0 < x ≤ x1
+    * log1pexp(x) ≈ x + exp(-x) for x1 < x ≤ x2
+    * log1pexp(x) ≈ x for x > x2
+=#
+@generated function _log1pexp_thresholds(::T) where {T<:Real}
```
tpapp, line 179: I am generally sympathetic to this approach, but would perhaps restrict to `T <: AbstractFloat`, to explicitly avoid types like `ForwardDiff.Dual`, etc, which are `<:Real`.

src/basicfuns.jl
tpapp: I would suggest just special-casing `ForwardDiff.Dual` to avoid a regression (ie define a method, use bounds for underlying type).

src/basicfuns.jl
tpapp: FWIW, I don't see the problem with `@generated`. @devmotion, can you please elaborate?

src/basicfuns.jl
cossio: Why avoid `ForwardDiff.Dual`? Note that `eps(ForwardDiff.Dual)` works fine here.

```
julia> eps(ForwardDiff.Dual(1.0,1.0))
2.220446049250313e-16
```

tpapp:
@cossio, @devmotion: sorry, I am a bit lost with all the comments, and I am not sure if there are any outstanding issues.

I think that the generalization of the branches based on the accuracy is a neat feature introduced in this PR. What's generally missing in Julia at the moment is an API to extract the accuracy/parent type of float-like types (there is `precision`, but it is not universally defined). I would just

1. make a `log1pexp_thresholds(::Type{T})` (note the type domain) part of the API, but not exported. Clarify what it is used for (current comments above the function are almost there) and mention that users can extend it in two ways: falling back to a type in `Base` (preferred), eg `log1pexp_thresholds(::Type{MyFloat{T}}) where T = log1pexp_thresholds(T)`, or defining their own.

2. define it falling back to the parent type of `ForwardDiff.Dual` using [Requires.jl](https://github.com/JuliaPackaging/Requires.jl), executed conditionally. Yes, this is a hack, and it should be moved to ForwardDiff in the future.

3. fall back to the precision bounds for `Float64` generally. This is suboptimal, but at least non-breaking. This could come with a deprecation warning (ie this method will be removed in the future).

How does this sound?

devmotion:
> define it falling back to the parent type of ForwardDiff.Dual using [Requires.jl](https://github.com/JuliaPackaging/Requires.jl), executed conditionally. Yes, this is a hack, and it should be moved to ForwardDiff in the future.

Requires is pretty bad, in particular if the conditional package is actually loaded the loading time is increased significantly usually (see e.g. https://github.com/JuliaLang/Pkg.jl/issues/1285#issue-476496324). I am strongly against adding a dependency on and using Requires here since it would mean enforcing it as an indirect dependency and making the package slower for central packages such as SpecialFunctions, StatsFuns, etc.

tpapp:
@devmotion: good point, then maybe 

1. fallback to `Float64` precision in this PR (not breaking anything), release, tagging a breaking change
2. get `ForwardDiff` to depend on this package and define a method
3. deprecate the fallback method.

cossio:
> * define it falling back to the parent type of `ForwardDiff.Dual` using [Requires.jl](https://github.com/JuliaPackaging/Requires.jl), executed conditionally. Yes, this is a hack, and it should be moved to ForwardDiff in the future.
> *

I don't get why this is needed. Can you post an example?

Note that

```
julia> ForwardDiff.derivative(log1pexp, 4.0)
0.9820137900379085
```

works (using this PR), and is using the Float64 thresholds.

src/basicfuns.jl
```diff
@@ -152,9 +152,38 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+function log1pexp(x::Real)
+    x0, x1, x2 = _log1pexp_thresholds(x)
+    if x ≤ x0
+        return exp(x)
+    elseif x ≤ x1
+        return log1p(exp(x))
+    elseif x ≤ x2
+        return x + exp(-x)
+    else
+        return oftype(x0, x)
```
devmotion, line 168: Not needed with the suggestion above:
```suggestion
        return x
```

src/basicfuns.jl
```diff
@@ -152,9 +152,38 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
 """
-log1pexp(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)
-log1pexp(x::Float32) = x < 9.0f0 ? log1p(exp(x)) : x < 16.0f0 ? x + exp(-x) : oftype(exp(-x), x)
+function log1pexp(x::Real)
+    x0, x1, x2 = _log1pexp_thresholds(x)
+    if x ≤ x0
+        return exp(x)
+    elseif x ≤ x1
+        return log1p(exp(x))
+    elseif x ≤ x2
```
devmotion, line 165: Julia uses 1-based indexing :slightly_smiling_face: 
```suggestion
    x1, x2, x3 = _log1pexp_thresholds(x)
    if x ≤ x1
        return exp(x)
    elseif x ≤ x2
        return log1p(exp(x))
    elseif x ≤ x3
```

devmotion: BTW is it easier to check `<` instead of `<=`? So maybe use
```suggestion
    x1, x2, x3 = _log1pexp_thresholds(x)
    if x < x1
        return exp(x)
    elseif x < x2
        return log1p(exp(x))
    elseif x < x3
```

src/basicfuns.jl
```diff
@@ -152,9 +152,38 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+    * log1pexp(x) ≈ x + exp(-x) for x1 < x ≤ x2
+    * log1pexp(x) ≈ x for x > x2
+=#
+@inline @generated function _log1pexp_thresholds(::T) where {T<:Real}
+    F = float(T)
+    ϵ = big(eps(F))
+    x0 = log(ϵ / 2)
+    x1 = -log(ϵ) / 2
+    x2 = -log(ϵ/2) - log(-log(ϵ/2)) - log(-log(ϵ/2)) / log(ϵ/2) # root of e^-x == x * ϵ/2
+    return F.((x0, x1, x2))
```
devmotion, line 185: With the suggestion above `F` is not needed. Additionally, always good to avoid unnecessary broadcasting operations:
```suggestion
    ϵ = big(eps(T))
    x0 = log(ϵ / 2)
    x1 = -log(ϵ) / 2
    x2 = -log(ϵ/2) - log(-log(ϵ/2)) - log(-log(ϵ/2)) / log(ϵ/2) # root of e^-x == x * ϵ/2
    return (T(x0), T(x1), T(x2))
```

devmotion (changes requested):

cossio:
So now all the thresholds are computed generically. There is a small change for `Float64` in that the original thresholds (currently in master and those in the paper) are: `(-37.0, 18.0, 33.0)`, whereas with the generic implementation in this PR they become: `(-36.7368005696771, 18.021826694558577, 33.231118823529634)`. Note that in the paper they are selected arbitrarily to make them round numbers, so I don't think this is a significant change. Though if we are very picky it could be slightly breaking.

If you do consider this breaking, I can fix this PR to make it fallback to the Maechler bounds (even for Float16 and BigFloat, which is not accurate, I have some examples of this). And then I can make another PR with the generic bound computation which would work for Float16, BigFloat, etc.

Honestly I wouldn't consider the change in bounds from `(-37.0, 18.0, 33.0)` to `(-36.73, 18.02, 33.23)` as significant, but I defer to your better judgement.

Here are updated benchmarks.

This PR:

```
julia> @benchmark log1pexp(x) setup=(x=30randn();) samples=10^7
BenchmarkTools.Trial: 263652 samples with 1000 evaluations.
 Range (min … max):   2.790 ns … 89.320 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     18.866 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   17.326 ns ±  8.482 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▄               ▃               █▁             ▄             
  █▂▁▁▁▁▁▁▁▁▁▁▁█▂▂█▃▂▂▁▁▁▁▁▁▁▁▁▁▁▂██▂▄▃▁▃▁▁▁▁▁▁▂▁█▁▇▁▃▁▃▁▁▃▁▁ ▂
  2.79 ns         Histogram: frequency by time        32.1 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

Current master:

```
julia> @benchmark log1pexp(x) setup=(x=30randn();) samples=10^7
BenchmarkTools.Trial: 219712 samples with 999 evaluations.
 Range (min … max):   9.339 ns … 132.006 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     20.783 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   20.694 ns ±   8.275 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

    █▄                 ▇ ▅            ▁  ▁                      
  ▇▇██▆▃▄▄▅▆▂▁▁▁▂▂▅▆█▂▇█▄█▃▂▁▂▁▄▄▂▅▂▅▂█▅▂█▃▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  9.34 ns         Histogram: frequency by time         43.2 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

src/basicfuns.jl
tpapp: FWIW, there is no indexing here, just notation. As long as it is documented, @cossio should just use whatever variables preferred as far as I am concerned.

tpapp:
> Though if we are very picky it could be slightly breaking.

I don't think they are, technically. If the calculations are correct, they should give the same end result, which is what the user should be concerned with.

Thanks for all the great work done in this PR. It turned out to be more involved than I expected from the first commit, and I appreciate that you saw it through.

src/basicfuns.jl
```diff
@@ -152,9 +152,38 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+        return x + exp(-x)
+    else
+        return oftype(x0, x)
+    end
+end
+
+#=
+returns a thresholds x0, x1, x2 such that:
+    * log1pexp(x) ≈ exp(x) for x ≤ x0
+    * log1pexp(x) ≈ log1p(exp(x)) for x0 < x ≤ x1
```
tpapp, line 175: Perhaps make a small comment about the tolerances for `≈` in this context.

src/basicfuns.jl
```diff
@@ -152,9 +152,50 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+    * log1pexp(x) ≈ x + exp(-x) for x1 < x ≤ x2
+    * log1pexp(x) ≈ x for x > x2
+=#
+@inline @generated function _log1pexp_thresholds(::T) where {T<:Real}
```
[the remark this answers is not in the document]
cossio: > Can you please benchmark the loading and compilation of LogExpFunctions and log1pexp, on master and with this change?

Sure, though I'm not sure how to do that :). How do you benchmark loading / compilation time of a package / function?
OTOH, unfortunately I don't see a nice way to avoid `@generated` here.

> Did you notice that the new thresholds are different for Float32?

They are slightly different. On master they are `(_, 9.0, 16.0)` (the first threshold is not used). After this PR,

```
_log1pexp_thresholds(::Float32) # (-16.635532f0, 7.9711924f0, 13.993f0)
```

> Additionally, I think it is better to help the compiler and hardcode the thresholds for `Float64`, `Float32`, etc. E.g., (not sure if dispatching with generated functions is problematic)

I don't see any issues, as far as I know generated functions dispatch fine. With `@generated` it's close to the hard-coded version:

```
julia> @code_typed LogExpFunctions._log1pexp_thresholds(1.0)
CodeInfo(
1 ─     return (-36.7368005696771, 18.021826694558577, 33.231118823529634)
) => Tuple{Float64, Float64, Float64}
```

src/basicfuns.jl
cossio: > Julia uses 1-based indexing slightly_smiling_face

Right, I was just being consistent with the notation of the paper. Since there is no name for the first bound I went with `x0`.

src/basicfuns.jl
cossio: Yep, done

src/basicfuns.jl
devmotion: Sure, I'm fine with the names, just thought it was a bit funny to point it out :smile: But if you keep them maybe change it to
```suggestion
    x0, x1, x2 = _log1pexp_thresholds(x)
    if x < x0
        return exp(x)
    elseif x < x1
        return log1p(exp(x))
    elseif x < x2
```
as it's a slightly simpler check in general (https://github.com/JuliaLang/julia/blob/5ac002038d93ea96f682bf005abc845427d5540c/base/operators.jl#L405)?

src/basicfuns.jl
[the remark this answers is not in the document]
cossio: Though hard-coding thresholds for Float32, Float64 is a way to make absolutely sure this PR is not breaking ...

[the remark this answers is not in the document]
devmotion: > Sure, though I'm not sure how to do that :). How do you benchmark loading / compilation time of a package / function?

Similar to e.g. https://github.com/JuliaDiff/ChainRulesCore.jl/pull/356#issuecomment-849956079 (and probably best to use `--startup-file=no` when starting Julia).

> I don't see any issues, as far as I know generated functions dispatch fine. With @generated it's close to the hard-coded version:

The issue is that the compiler has to work out the values on it's own - which increases compilation time. If you provide them, that's not needed for these common types.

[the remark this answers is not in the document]
devmotion: I'm fine with slightly different bounds (would be nice to make a proper analysis for how close the values are in this range but they seem sufficiently close), the main motivation is to avoid having to compile the generated function for these types.

src/basicfuns.jl
cossio: Thanks I added the suggested optimizations on the computation of the last threshold. I prefer to keep x0, x1, x2 names though since that's what's used in the paper.

src/basicfuns.jl
```diff
@@ -152,9 +152,41 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+function log1pexp(x::Real)
+    x0, x1, x2 = _log1pexp_thresholds(x)
+    if x ≤ x0
+        return exp(x)
+    elseif x ≤ x1
+        return log1p(exp(x))
+    elseif x ≤ x2
+        return x + exp(-x)
+    else
+        return float(x)
```
devmotion, line 168: It's simpler to just compute `float(x)` upfront - this will get rid of all promotions in checks and computations. Just use
```suggestion
        return x
```
and apply https://github.com/JuliaStats/LogExpFunctions.jl/pull/37/files#r821736843 :slightly_smiling_face:

src/basicfuns.jl
cossio: Sure, I replaced <= with <

src/basicfuns.jl
```diff
@@ -152,9 +152,50 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
 
 This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref).
+
+See:
+ * Martin Maechler (2012) [“Accurately Computing log(1 − exp(− |a|))”](http://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf)
 """
-log1pexp(x::Real) = x < 18.0 ? log1p(exp(x)) : x < 33.3 ? x + exp(-x) : oftype(exp(-x), x)
-log1pexp(x::Float32) = x < 9.0f0 ? log1p(exp(x)) : x < 16.0f0 ? x + exp(-x) : oftype(exp(-x), x)
+function log1pexp(x::Real)
```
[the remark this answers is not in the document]
cossio: > We can avoid the promotions in the conditionals by using

Ok I incorporated this change too.

src/basicfuns.jl
cossio: Yep, done

src/basicfuns.jl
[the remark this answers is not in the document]
cossio: Ok I introduced the hard-coded thresholds for Float32, Float64, with a comment explaining why. To be extra safe, I also set them to the values they had before this PR for those two cases.

src/basicfuns.jl
cossio: I'm marking this as resolved since it refers to an older version of the code.

src/basicfuns.jl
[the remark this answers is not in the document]
cossio: Marking resolved (older version of code)

src/basicfuns.jl
```diff
@@ -152,9 +152,47 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+
+where the tolerances of the approximations ≈ are on the order of eps(T).
+=#
+@inline @generated function _log1pexp_thresholds(::T) where {T<:Real}
+    ϵ = big(eps(T))
+    x0 = log(ϵ / 2)
+    x1 = -log(ϵ) / 2
+    x2 = -x0 - log(-x0) * (1 + 1 / x0) # ≈ root of e^-x == x * ϵ/2 via asymptotics of Lambert's W function
+    return (T(x0), T(x1), T(x2))
+end
```
devmotion, line 189: One could also use a `if @generated` block to let the compiler decide if it uses the `@generated` version or not (https://docs.julialang.org/en/v1/manual/metaprogramming/#Optionally-generated-functions).

devmotion (commented):
I think it looks quite good. My largest remaining concern is that basing the thresholds on the type is invalid for some number types - e.g., for `BigFloat`s always a precision of 256 will be used, regardless of the input. I'm not sure how to avoid this - as usual, the value provides more information than the type but a regular function would mean that the thresholds are recomputed every time the function is called which would destroy any performance benefits. On the other hand, it is not guaranteed that this does not happen with the generated function: The values might still be recomputed every time the function is called (as explained in https://docs.julialang.org/en/v1/manual/metaprogramming/#Generated-functions). I'm not sure what's the best way to deal with this, maybe we should at least add a special definition of `log1pexp` (possibly just based on `log(exp(-abs(x)))`) for `BigFloat`. Maybe it would also be helpful to use `precision` to avoid compiling and recomputing thresholds for different number types with the same precision?

devmotion:
Maybe one could check `precision` in the default definition, and use (hardcoded) thresholds for the precisons of Float64, Float32, and Float16, and otherwise just use `log(exp(-abs(x)) + (x < zero(x) : zero(x) : x)` (of course, in a nicer way)?

src/basicfuns.jl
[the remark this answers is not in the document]
cossio: > So it's not clear that the values are only computed once, they might be recomputed the next time the function is called.

In this case you can see from `@code_typed` that the thresholds are cached fine.

src/basicfuns.jl
```diff
@@ -152,9 +152,47 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+#= For Float64, Float32 we can hard-code the thresholds to (hopefully) reduce compilation
+times. We will use here the exact same thresholds that were used before the function
```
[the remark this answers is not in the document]
cossio: Ok done.

src/basicfuns.jl
cossio: > One could also use a `if @generated` block to let the compiler decide if it uses the `@generated` version or not (https://docs.julialang.org/en/v1/manual/metaprogramming/#Optionally-generated-functions).

What's the benefit of this?

devmotion: It can reduce compile time cost:
> Generated functions can achieve high efficiency at run time, but come with a compile time cost: a new function body must be generated for every combination of concrete argument types. Typically, Julia is able to compile "generic" versions of functions that will work for any arguments, but with generated functions this is impossible. This means that programs making heavy use of generated functions might be impossible to statically compile.
>
> To solve this problem, the language provides syntax for writing normal, non-generated alternative implementations of generated functions.

src/basicfuns.jl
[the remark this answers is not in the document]
devmotion: You can't draw any such conclusions from `@code_typed`. It only shows the result of code generation but does not tell you if it will be recomputed or cached. And even if it is fine for your specific Julia version for Float64, according to the docs you should not rely on that it's cached in the same way for other types and/or Julia versions.

cossio:
> Maybe one could check `precision` in the default definition, and use (hardcoded) thresholds for the precisons of Float64, Float32, and Float16, and otherwise just use `log(exp(-abs(x)) + (x < zero(x) : zero(x) : x)` (of course, in a nicer way)?

`precision` is not defined for `Dual`

devmotion:
That's why I suggested basing it on the precision instead of types - but if you make the values consistent types should not be a problem either.

There's always a compile time cost with `@generated` functions and BigFloat computations, so we should really hardcode it for common types/precisions.

The main problem with types is really just that they do not provide enough information in general.

cossio:
Is this a real issue?

```julia
julia> precision(exp(BigFloat(1.0; precision=500)))
256
```

It seems `BigFloat` computations always return precision 256 regardless of input.

cossio:
@devmotion I think functions in Base taking `x::BigFloat` inputs work with precision given by the type `precision(BigFloat)`, not  the value `precision(x)`. See the `exp` example above. 

By having the threholds generated in this PR, this means the thresholds will be computed on the value of `precision(BigFloat)` at the first time `log1pexp_thresholds` is called. So if one latter does `setprecision(BigFloat, prec)`, it won't work fine.

That's the only issue I see at the moment.

Regarding compilation time, if I remove the hard-coded bounds and rely exclusively on the generated function, I don't see any compilation latency.

```
julia> @time LogExpFunctions.log1pexp(1.0) # first call on a fresh session
  0.000001 seconds
```

But maybe there are better ways of measuring that.

devmotion:
> precision is not defined for Dual

Maybe it should be added. IMO this would be reasonable since e.g. `eps` is defined already.

> I think functions in Base taking x::BigFloat inputs work with precision given by the type precision(BigFloat), not the value precision(x). See the exp example above.
> 
> By having the threholds generated in this PR, this means the thresholds will be computed on the value of precision(BigFloat) at the first time log1pexp_thresholds is called. So if one latter does setprecision(BigFloat, prec), it won't work fine.

Exactly, that's what I mean. And it is easy to come up with example where this will cause incorrect results:
```julia
julia> using LogExpFunctions

julia> f(x) = log1pexp(x) == log1p(exp(x));

julia> setprecision(128) do
           f(big"100.0")
       end
true

julia> setprecision(512) do
           f(big"100.0")
       end
false
```
whereas a value-based non-`@generated` implementation of `_log1pexp_thresholds` correctly returns
```julia
julia> setprecision(128) do
           f(big"100.0")
       end
true

julia> setprecision(512) do
           f(big"100.0")
       end
true
```
Of course, `BigFloat` could be fixed e.g. (with a performance regression) by just implementing `log1pexp(::BigFloat)`. But this does not solve the general problem - fir any other `Real` number type without fixed type-encoded precision `log1pexp` will also return incorrect results.

> Regarding compilation time, if I remove the hard-coded bounds and rely exclusively on the generated function, I don't see any compilation latency.

I think that's difficult to measure. However, we know for sure that
1. generated functions have an increased compilation time in general (even if it is small) and has to perform calculations with `BigFloat` (which is not instantaneous either), and
2. it is not guaranteed that generated functions cache the result of their code generation.

Both things can be fixed easily by hardcoding the results of the code generation, so it feels a bit strange to not do it.

cossio:
> Both things can be fixed easily by hardcoding the results of the code generation, so it feels a bit strange to not do it.

Hmmm I have to think about it. The idea here was that the generated code worked with everything (BigFloat, Dual, Float16, etc.), up to the setprecision issue with BigFloat. I can hardcode Float64, Float32, etc, but then you have to handle also Dual? I'm just not sure how it would work.

cossio:
> Maybe it should be added. IMO this would be reasonable since e.g. `eps` is defined already.

Yeah, so `precision(::Dual)` should be defined in `ForwardDiff.jl, right?

devmotion:
I realized that we don't need a generated function. Instead we could use
```julia
import IrrationalConstants

function _log1pexp_thresholds(x::Real)
    z = (precision(x) - 1) / log2(oftype(x, ℯ)) # define an irrational constant for `log2(ℯ)`?
    x0 = - z - IrrationalConstants.logtwo
    x1 = z / 2
    x2 = -x0 - log(-x0) * (1 + 1 / x0)
    return (x0, x1, x2)
end
```
The compiler can optimize this quite well (still, hardcoding Float64/32/16 would mean less work for these types):
```julia
julia> @code_typed _log1pexp_thresholds(0.0)
CodeInfo(
1 ─     return (-36.7368005696771, 18.021826694558577, 33.23111882352963)
) => Tuple{Float64, Float64, Float64}
```
I compared it with the non-generated function
```julia
function _log1pexp_thresholds2(::T) where {T<:Real}
    ϵ = big(eps(T))
    x0 = log(ϵ / 2)
    x1 = -log(ϵ) / 2
    x2 = -x0 - log(-x0) * (1 + 1 / x0) # ≈ root of e^-x == x * ϵ/2 via asymptotics of Lambert's W function
    return (T(x0), T(x1), T(x2))
end
```
to get an estimate of how much time it takes to evaluate both functions.
```julia
julia> @btime _log1pexp_thresholds($1.0);
  0.020 ns (0 allocations: 0 bytes)

julia> @btime _log1pexp_thresholds2($1.0);
  18.648 μs (29 allocations: 1.52 KiB)
```
As expected the computation with `BigFloat`s is much slower - and additionally depends on the chosen precision which is another issue that affects the thresholds:
```julia
julia> _log1pexp_thresholds(1.0)
(-36.7368005696771, 18.021826694558577, 33.23111882352963)

julia> setprecision(10) do
           _log1pexp_thresholds2(1.0)
       end
(-36.75, 18.03125, 33.25)

julia> setprecision(100) do
           _log1pexp_thresholds2(1.0)
       end
(-36.7368005696771, 18.021826694558577, 33.231118823529634)
```

So I think we could
- replace the generated function with the `precision`-based alternative above
- define hard-coded thresholds for `Float64`/`Float32`/`Float16` to a) reduce the compilation time and more importantly to b) be absolutely certain that the function is optimized away and not recomputed in every function call
- define `precision` for `Dual` in ForwardDiff

Since we only work with `x = float(_x)`, i.e., floating point numbers `precision(x)` should be defined for all inputs. However, I guess some downstream packages (such as ForwardDiff) are missing it and would have to be fixed.

devmotion:
> Yeah, so `precision(::Dual)` should be defined in `ForwardDiff.jl`, right?

One has to take care of the `base` keyword argument in Julia 1.8 as well and support both values and types as inputs (since for variable precision type they are different). So one should add something like
```julia
# overload Base._precision to support the base keyword in Julia 1.8
let precision = isdefined(Base, :_precision) ? :_precision : :precision
    @eval begin
        Base.$precision(d::Dual) = Base.$precision(value(d))
        Base.$precision(::Type{D}) where {D<:Dual} = Base.$precision(valtype(D))
    end
end
```
I'll make a PR.

cossio:
Nice!

> * define hard-coded thresholds for `Float64`/`Float32`/`Float16` to a) reduce the compilation time and more importantly to b) be absolutely certain that the function is optimized away and not recomputed in every function call

I only note that these hard-coded versions for `::Float64` etc. won't apply for e.g. `Dual{Float64}`. Do you have any comment on that? As far as I can tell `_log1pexp_thresholds(x::Dual{Float64})` will just fallback to the generic implementation. So I guess for consistency it'd be best to make sure the same thresholds are used in both cases.

devmotion:
https://github.com/JuliaDiff/ForwardDiff.jl/pull/580

cossio:
In terms of compat: How can we rely here on `precision(::Dual)` being defined?

devmotion:
> I only note that these hard-coded versions for `::Float64` etc. won't apply for e.g. `Dual{Float64}`. Do you have any comment on that? As far as I can tell `_log1pexp_thresholds(x::Dual{Float64})` will just fallback to the generic implementation. So I guess for consistency it'd be best to make sure the same thresholds are used in both cases.

Actually, `log1pexp(::Dual)` will always use the `Float64` bounds, regardless of how we define them (if the primal in the dual number is of type `Float64`). We also don't even need the `precision` PR, even though it is useful :smile: In ForwardDiff (and e.g. also in Tracker) `log1pexp` is defined for `Dual` (and `TrackedReal`) via DiffRules in more recent versions of DiffRules: https://github.com/JuliaDiff/ForwardDiff.jl/blob/eed0ff37de790572440691a3b998bcc68d834875/src/dual.jl#L198-L212 and https://github.com/JuliaDiff/ForwardDiff.jl/blob/eed0ff37de790572440691a3b998bcc68d834875/src/dual.jl#L395-L409

src/basicfuns.jl
```diff
@@ -152,9 +152,62 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+    elseif x < x2
+        return x + exp(-x)
+    else
+        return x
+    end
+end
+
+#= The precision of BigFloat cannot be computed from the type only.
+Since computing the dynamically thresholds is slow, prefer to use
+a slower but accurate implementation in this case. =#
```
devmotion, line 175: This seems contradicting - if the function is slow, we should use a faster alternative but not a slower one? Did you benchmark both approaches?

src/basicfuns.jl
```diff
@@ -152,9 +152,62 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+    * log1pexp(x) ≈ x for x > x2
+
+where the tolerances of the approximations are on the order of eps(typeof(x)).
+For types for which `precision(x)` depends only on the type of `x`, the compiler
+should optimize away all computations done here.
+=#
+@inline function _log1pexp_thresholds(x::Real)
+    z = (precision(x) - 1) * oftype(float(x), IrrationalConstants.logtwo)
+    x0 = -z - IrrationalConstants.logtwo
+    x1 = z / 2
```
devmotion, line 199: We don't need `float(x)`, the function is only called with floating point number types. One could also avoid the subtraction:
```suggestion
    prec = precision(x)
    logtwo = oftype(x, IrrationalConstants.logtwo)
    x0 = - prec * logtwo
    x1 = ((prec - 1) * logtwo) / 2
```

src/basicfuns.jl
cossio: What I meant is that for BigFloat, the approach with the thresholds is slower than this special-cased method. Here are the benchmarks I run.

Method specialized for `BigFloat` without thresholds.

```julia
julia> @benchmark log1pexp(big(x)) setup=(x=30randn();) samples=10^7
BenchmarkTools.Trial: 58177 samples with 7 evaluations.
 Range (min … max):   4.600 μs … 759.421 μs  ┊ GC (min … max): 0.00% … 94.03%
 Time  (median):      8.594 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   11.937 μs ±   8.873 μs  ┊ GC (mean ± σ):  0.41% ±  0.97%

    ▃██▃                                                        
  ▁▄████▆▃▂▁▁▁▁▁▁▁▁▂▄▆▆▆▅▃▂▂▁▁▁▁▁▁▁▁▁▁▁▁▂▃▅▄▄▄▄▆▃▂▂▂▁▁▁▁▁▁▁▁▁▁ ▂
  4.6 μs          Histogram: frequency by time         27.4 μs <

 Memory estimate: 504 bytes, allocs estimate: 11.
```

Method with thresholds.

```
julia> @benchmark log1pexp(big(x)) setup=(x=30randn();) samples=10^7
BenchmarkTools.Trial: 203069 samples with 1 evaluation.
 Range (min … max):  12.555 μs …  4.530 ms  ┊ GC (min … max): 0.00% … 98.47%
 Time  (median):     21.679 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   22.388 μs ± 26.404 μs  ┊ GC (mean ± σ):  0.70% ±  0.62%

                    ▄▂█ ▂ ▃                                    
  ▁▁▁▂▅▇▇▆▅▄▃▂▂▁▁▁▁▃█████▆█▆▄▃▂▂▁▁▁▁▁▁▁▂▃▂▃▃▂▃▂▂▃▂▂▁▂▁▁▁▁▁▁▁▁ ▂
  12.6 μs         Histogram: frequency by time        38.5 μs <

 Memory estimate: 1.95 KiB, allocs estimate: 38.
```

 I'll rewrite the comment.

src/basicfuns.jl
cossio: Agree with the removing `float(x)`. For the computations, it seems you are eliminating a subtraction but introducing more calculations in `x0, x1`.

devmotion: Yeah. Ideally, there would be a constant for `log(sqrt(2))` that we could use, then it would lead to the much simpler code
```suggestion
    prec = precision(x)
    x0 = - prec * oftype(x, IrrationalConstants, logtwo)
    x1 = (prec - 1) * oftype(x, IrrationalConstants.logsqrt2)
```

cossio:
Unfortunately this version:

```julia
@inline function _log1pexp_thresholds(x::Real)
    prec = precision(x)
    logtwo = oftype(x, IrrationalConstants.logtwo)
    x0 = -prec * logtwo
    x1 = (prec - 1) * logtwo / 2
    x2 = -x0 - log(-x0) * (1 + 1 / x0) # approximate root of e^-x == x * ϵ/2 via asymptotics of Lambert's W function
    return (x0, x1, x2)
end
```

is not compiled away in Julia 1.0. Actually `@code_typed` shows a lot of generated code. Julia 1.6 is much better but still is not completely elided. Amazing how Julia 1.7 kills it :smile: 

This supports having the hard-coded thresholds for `Float`s. Not sure if we should worry about regressions on Julia 1.6 code though.

src/basicfuns.jl
```diff
@@ -152,9 +152,65 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+
+#= The precision of BigFloat cannot be computed from the type only and computing
+thresholds is slow. Therefore prefer version without thresholds in this case. =#
+function log1pexp(x::BigFloat)
+    if x > 0
+        return x + log1p(exp(-x))
+    else
+        return log1p(exp(x))
+    end
+end
```
devmotion, line 181: Maybe in one line?
```suggestion
log1pexp(x::BigFloat) = x > 0 ? x + log1p(exp(-x)) : log1p(exp(x))
```

src/basicfuns.jl
```diff
@@ -152,9 +152,65 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+    prec = precision(x)
+    logtwo = oftype(x, IrrationalConstants.logtwo)
+    x0 = -prec * logtwo
+    x1 = (prec - 1) * logtwo / 2
+    x2 = -x0 - log(-x0) * (1 + 1 / x0) # approximate root of e^-x == x * ϵ/2 via asymptotics of Lambert's W function
+    return (x0, x1, x2)
+end
+
+#= For Float64, Float32 we can hard-code the thresholds to make absolutely sure they are not
+recompued each time. Also, _log1pexp_thresholds is not completely elided by the
```
devmotion, line 205:
```suggestion
recomputed each time. Also, _log1pexp_thresholds is not completely elided by the
```

src/basicfuns.jl
```diff
@@ -152,9 +152,65 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+compiler in Julia 1.0 / 1.6 which LogExpFunctions intends to support.
+For Float64 we use the same exact thresholds given by Maechler 2012,
+since these were the ones used before _log1pexp_thresholds was introduced (which outputs
+close but not identical thresholds), to reduce any (small) risk of breakage.
+For Float32 and Float16 we use truncated versions of the output of _log1pexp_thresholds. =#
```
devmotion, line 210: Maybe it would be interesting to add the precision-based values as a comment?

cossio: Not sure what you mean?

devmotion: I mean the values you would get from the non-hardcoded version. Then it would be easier to see how it's rounded.

test/basicfuns.jl
```diff
@@ -110,15 +110,30 @@ end
...
+    # large arguments
+    @test (@inferred log1pexp(1e4)) ≈ 1e4
+    @test (@inferred log1pexp(1f4)) ≈ 1f4
+    @test iszero(@inferred log1pexp(-1e4))
+    @test iszero(@inferred log1pexp(-1f4))
+
+    # compare to accurate but slower implementation
+    correct_log1pexp(x::Real) = x > 0 ? x + log1p(exp(-x)) : log1p(exp(x))
+    # large range needed to cover all branches, for all floats (from Float16 to BigFloat)
+    @testset "log1pexp($x::$T)" for T in (Float16, Float32, Float64, BigFloat), x in -300:300
```
devmotion, line 134: Maybe test also different precisions for BigFloat?

src/basicfuns.jl
cossio: Ah yes, ok.

test/basicfuns.jl
cossio: Yep, done. Added tests for precisions 10, 20, 50. 100.

src/basicfuns.jl
cossio: Ended up liking this version since the subtraction is between ints which is exact. Marking resolved.

src/basicfuns.jl
cossio: Marking resolved.

src/basicfuns.jl
```diff
@@ -152,9 +152,61 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+end
+
+#=
+For common types we hard-code the thresholds to make absolutely sure they are not
+recomputed each time. Also, _log1pexp_thresholds is not completely elided by the
+compiler in Julia 1.0 / 1.6. We round the outputs of the non-hardcoded version:
+
+* _log1pexp_thresholds(::Float64) = (-36.7368005696771, 18.021826694558577, 33.23111882352963)
+* _log1pexp_thresholds(::Float32) = (-16.635532f0, 7.9711924f0, 13.993f0)
+* _log1pexp_thresholds(::Float16) = (-7.625, 3.467, 5.86)
```
devmotion, line 205: I guess for `Float16` we could use these values? Or is there any to reason to truncate them or not to rely on these values?

src/basicfuns.jl
```diff
@@ -152,9 +152,61 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
...
+    x2 = -x0 - log(-x0) * (1 + 1 / x0) # approximate root of e^-x == x * ϵ/2 via asymptotics of Lambert's W function
+    return (x0, x1, x2)
+end
+
+#=
+For common types we hard-code the thresholds to make absolutely sure they are not
+recomputed each time. Also, _log1pexp_thresholds is not completely elided by the
+compiler in Julia 1.0 / 1.6. We round the outputs of the non-hardcoded version:
+
+* _log1pexp_thresholds(::Float64) = (-36.7368005696771, 18.021826694558577, 33.23111882352963)
```
devmotion, line 203: Any explanation for why Maechler finds the roots `-36.39022698`, `16.40822612`, and `33.2783501`?

devmotion (approved):
I think this looks good, I have only a question about the `Float64` and `Float16` values left (and I think it would also be fine to make the hardcoded values consistent with the generic fallback definition, but I don't have a strong opinion there).

In general, now I think the PR should be an improvement for all inputs `x` where `float(x)` is a fixed-precision number since then (at least in recent Julia versions) the compiler can optimize away thresholds even for non-standard types. It's only a bit problematic for variable precision numbers - in contrast to the `@generated` version the results will be correct but since the thresholds have to be recomputed every time it might cause performance regressions. IMO this is a bit annoying but much better than silently returning wrong results, and it can be fixed in the same way as done here for `BigFloat`.

Can you update the version number as well?

src/basicfuns.jl
cossio: What do you mean?

cossio: Ah you mean on the paper. Maechler is finding the thresholds numerically by calling a root search routine (you have to find where the approximation first "intersects" the exact expression at each precision). This is much slower than what I wrote here. I am just computing an approximation to these roots, see https://github.com/JuliaStats/LogExpFunctions.jl/pull/37#issuecomment-1061731559.

src/basicfuns.jl
cossio: Mmm no. I was just truncating them for style. But I also think it's more consistent to just use the same thresholds given by the generic fallback, so I'll do that.

cossio:
@devmotion I attach here the calculation of the thresholds I am using in this PR.

[log1pexp.pdf](https://github.com/JuliaStats/LogExpFunctions.jl/files/8218470/log1pexp.pdf)

devmotion:
It seems the approximations in your notes are slightly different from the values in the PR?
```julia
julia> log(eps(Float64))
-36.04365338911715

julia> -log(2*eps(Float64)) / 2
17.675253104278607
```

cossio:
The epsilon in the notes is `eps(T) / 2`, to ensure correct rounding to the nearest float.

cossio:
Merge?
Or are we waiting for https://github.com/JuliaDiff/ForwardDiff.jl/pull/580 first?

tpapp:
My understanding is that the current solution is broken without that PR, so for now I would suggest waiting for that.

@cossio, in the meantime, can you please add a `@test_broken` for `ForwardDiff.Dual`s to verify this? We can change that simply when the related PR is merged.

tpapp:
@devmotion: thanks for the clarification. In that case, please feel free to merge.

PR #38, opened 2022-03-11
logsumexp!(r, X)

### LINKED ISSUE #14: Add `logsumexp!(r, X)` (which behaves like sum!)
It would be nice to have `logsumexp!(r, X)`, which sums `X` over the singleton dimensions of `r` and writes the results in `r` (like `sum!(r, X)`)

cossio:
Add `logsumexp!(r, X)`, in-place version of `logsumexp`.

From the docstring:

```
Compute `log.(sum!(r, exp.(X); dims=dims))`, summing over the singleton dimensions of `r`,
and overwriting `r` with the results.
The calculation is done in a numerically stable way that avoids intermediate
over- and underflow.
```

Close #14 .

### DIFF
--- a/src/logsumexp.jl
+++ b/src/logsumexp.jl
@@ -126,3 +126,17 @@ function _logsumexp_onepass_op(xmax1::T, xmax2::T, r1::R, r2::R) where {T<:Numbe
     end
     return xmax, r
 end
+
+"""
+$(SIGNATURES)
+
+Compute `log.(sum!(r, exp.(X); dims=dims))`, summing over the singleton dimensions of `r`,
+and overwriting `r` with the results.
+The calculation is done in a numerically stable way that avoids intermediate
+over- and underflow.
+"""
+function logsumexp!(r::AbstractArray, X::AbstractArray)
+	m = maximum!(similar(r), X)
+	sum!(r, exp.(X .- m))
+	return r .= log.(r) .+ m
+end
--- a/src/logsumexp.jl
+++ b/src/logsumexp.jl
@@ -136,7 +136,7 @@ The calculation is done in a numerically stable way that avoids intermediate
 over- and underflow.
 """
 function logsumexp!(r::AbstractArray, X::AbstractArray)
-	m = maximum!(similar(r), X)
-	sum!(r, exp.(X .- m))
-	return r .= log.(r) .+ m
+    m = maximum!(similar(r), X)
+    sum!(r, exp.(X .- m))
+    return r .= log.(r) .+ m
 end
--- a/src/logsumexp.jl
+++ b/src/logsumexp.jl
@@ -137,6 +137,7 @@ over- and underflow.
 """
 function logsumexp!(r::AbstractArray, X::AbstractArray)
     m = maximum!(similar(r), X)
-    sum!(r, exp.(X .- m))
-    return r .= log.(r) .+ m
+    c = sum!(similar(r), X .== m)
+    sum!(r, (X .< m) .* exp.(X .- m))
+    return r .= log.(c) .+ log1p.(r ./ c) .+ m
 end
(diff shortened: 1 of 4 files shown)


cossio:
CI failure on Julia nightly is unrelated to this PR. Upstream issue: https://github.com/JuliaCI/julia-buildkite/pull/59

cossio:
Right now I'm don't know how to implement an in-place version of the online algorithm. Can we leave that for a future PR? We can then use `logsumexp!(r, X)` given here as a baseline benchmark to see if the online version is faster also for the in-place method.

Regarding `log1p`. I'm not sure it's needed here. Can you expand?

devmotion:
We should use the out-of-place version as baseline. IMO adding an in-place version only makes sense if 1) it has (at least) the same accuracy and 2) is not slower. I am not sure if the current draft satisfies these goals.

test/basicfuns.jl
```diff
@@ -266,6 +266,13 @@ end
     @test @inferred(logsumexp(x for x in xs)) == logsumexp(xs)
 end
 
+@testset "logsumexp!" begin
+    X = randn(4,3,2)
+    r = randn(1,3,2)
+    @inferred logsumexp!(r, X)
+    @test r ≈ logsumexp(X; dims=1)
```
devmotion, line 273: This requires many more tests, similar to the tests for `logsumexp`.

Regarding underflow: I just checked and indeed the current draft is significantly less accurate than `logsumexp`. Whereas `logsumexp([1e-20, log(1e-20)]) ≈ 2e-20` do we get here
```julia
julia> X = [1e-20, log(1e-20)];

julia> r = [1.3];

julia> logsumexp!(r, X)
1-element Vector{Float64}:
 1.0e-20
```

cossio: Good catch.

src/logsumexp.jl
```diff
@@ -126,3 +126,18 @@ function _logsumexp_onepass_op(xmax1::T, xmax2::T, r1::R, r2::R) where {T<:Numbe
...
+Compute `log.(sum!(r, exp.(X); dims=dims))`, summing over the singleton dimensions of `r`,
+and overwriting `r` with the results.
+The calculation is done in a numerically stable way that avoids intermediate
+over- and underflow.
+"""
+function logsumexp!(r::AbstractArray, X::AbstractArray)
+    m = maximum!(similar(r), X)
+    c = sum!(similar(r), X .== m)
+    sum!(r, (X .< m) .* exp.(X .- m))
+    return r .= log.(c) .+ log1p.(r ./ c) .+ m
```
devmotion, line 142: Can you benchmark the method? There are so many additional allocations that intuitively I assume it will be slower than the out-of-place method for a wide range of sizes of `X` and `r`.

cossio: Yeah it is slower. I'll post some benchmarks for reference.

cossio:
So far in-place version of this PR is slower than the out-of-place online algorithm already in master.

```
julia> @btime logsumexp(X; dims=1) setup=(X=500 .* randn(100_000, 100);); # master
  122.210 ms (2 allocations: 2.64 KiB)

julia> @btime logsumexp(X; dims=2) setup=(X=500 .* randn(100_000, 100);); # master
  154.873 ms (8 allocations: 2.29 MiB)

julia> @btime logsumexp!(r, X) setup=(r=randn(1,100); X=500 .* randn(100_000, 100);); # this PR
  183.793 ms (8 allocations: 77.49 MiB)

julia> @btime logsumexp!(r, X) setup=(r=randn(100_000, 1); X=500 .* randn(100_000, 100);); # this PR
  232.209 ms (10 allocations: 79.02 MiB)
```

devmotion:
I put together a PR that is based on the one-pass algorithm: https://github.com/JuliaStats/LogExpFunctions.jl/pull/39

It seems to work but I still have to add tests. I plan to finish it this weekend.

cossio:
Closing in favor of #39

PR #39, opened 2022-03-12
Add `logsumexp!`

devmotion:
Alternative to https://github.com/JuliaStats/LogExpFunctions.jl/pull/38.

```julia
julia> using LogExpFunctions, BenchmarkTools

julia> @btime logsumexp(X; dims=1) setup=(X=500 .* randn(100_000, 100);); # master
  86.982 ms (2 allocations: 2.64 KiB)

julia> @btime logsumexp(X; dims=2) setup=(X=500 .* randn(100_000, 100);); # master
  109.627 ms (8 allocations: 2.29 MiB)

julia> @btime logsumexp!(r, X) setup=(r=randn(1,100); X=500 .* randn(100_000, 100);); # this PR
  85.421 ms (1 allocation: 1.77 KiB)

julia> @btime logsumexp!(r, X) setup=(r=randn(100_000, 1); X=500 .* randn(100_000, 100);); # this PR
  110.363 ms (2 allocations: 1.53 MiB)
```

~~Needs tests.~~
I added tests.

### DIFF
--- a/src/logsumexp.jl
+++ b/src/logsumexp.jl
@@ -10,6 +10,8 @@ pass over the data.
 # References
 
 [Sebastian Nowozin: Streaming Log-sum-exp Computation](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
+
+See also [`logsumexp!`]
 """
 logsumexp(X) = _logsumexp_onepass(X)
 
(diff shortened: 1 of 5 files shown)


src/logsumexp.jl
```diff
@@ -24,9 +26,32 @@ The result is computed using a single pass over the data.
...
 """
 logsumexp(X::AbstractArray{<:Number}; dims=:) = _logsumexp(X, dims)
 
+"""
+$(SIGNATURES)
+
+Compute `out .= log.(sum!(out, exp.(X)))` in a numerically stable way that avoids
+intermediate over- and underflow.
+
+The result is computed using a single pass over the data.
```
cossio, line 40:
```suggestion
Compute `log.(sum!(r, exp.(X); dims=dims))`, summing over the singleton dimensions of `r`,
and overwriting `r` with the results.

The result is computed using a single pass over the data, in a numerically stable way that avoids intermediate
over- and underflow.
```

I think the `out .= ...` part is confusing. It's more clear to say explicitly that `r` is overwritten.
Also the docstring should mention that the sum is carried out over singleton dimensions of `r`.
Compare to the docstring of `sum!`.

cossio:
Thanks!

src/logsumexp.jl
devmotion: > Also the docstring should mention that the sum is carried out over singleton dimensions of out.

That's already implied by `sum!` (which BTW does not support a `dims` keyword argument).

devmotion: I updated the docstring.

src/logsumexp.jl
```diff
@@ -24,9 +26,31 @@ The result is computed using a single pass over the data.
...
+
+Compute [`logsumexp`](@ref) of `X` over the singleton dimensions of `out`, and write results to `out`.
+
+The result is computed in a numerically stable way that avoids intermediate over- and underflow using a single pass over the data.
+
+# References
+
+[Sebastian Nowozin: Streaming Log-sum-exp Computation](http://www.nowozin.net/sebastian/blog/streaming-log-sum-exp-computation.html)
+
+See also [`logsumexp`](@ref)
```
tpapp, line 45:
```suggestion
See also [`logsumexp`](@ref).
```

tpapp (approved):
LGTM, just suggested a minor punctuation change.

src/logsumexp.jl
devmotion: I fixed it.

PR #40, opened 2022-03-13
Fix `logsumexp!` with output arrays of abstract `eltype`

devmotion:
This PR fixes
```julia
julia> out = [missing, 1.0];

julia> x = rand(2, 4);

julia> logsumexp!(out, x)
ERROR: MethodError: no method matching logsumexp!(::Vector{Union{Missing, Float64}}, ::Matrix{Float64})
Closest candidates are:
  logsumexp!(::AbstractArray{<:Number}, ::AbstractArray{<:Number}) at ~/.julia/dev/LogExpFunctions/src/logsumexp.jl:45
Stacktrace:
 [1] top-level scope
   @ REPL[8]:1
```
and adds support for output arrays with abstract `eltype`s, similar to e.g. `sum!`:
```julia
julia> sum!(out, x)
2-element Vector{Union{Missing, Float64}}:
 1.2939717364740972
 2.348547173687184
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LogExpFunctions"
 uuid = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 authors = ["StatsFun.jl contributors, Tamas K. Papp <tkpapp@gmail.com>"]
-version = "0.3.9"
+version = "0.3.10"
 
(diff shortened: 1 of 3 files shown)


tpapp (approved):
LGTM.

PR #41, opened 2022-04-09
new implementation of logaddexp

lucasondel:
This pull request proposes a new implementation of `logaddexp` which is much faster than the current one.

Here is a small benchmark to illustrate the benefit:
```
julia> using LogExpFunctions
julia> using BenchmarkTools
julia> function logaddexp_new(x, y)
         if isnan(x) return x end
             if isnan(y) return y end
  
             diff = zero(promote_type(typeof(x), typeof(y)))
             if x < y
                 diff = x - y
                 x = y
             else
                 diff = y - x
             end
  
             if diff >= log(eps())
                 return x + log1p(exp(diff))
             else
                 return x
             end
       end;
julia> function logaddexp_old(x, y)
         Δ = x == y ? zero(x - y) : abs(x - y)
         return max(x, y) + log1pexp(-Δ)                                                              
       end;
julia> x = 1000*randn(100000);
julia> y = 1000*randn(100000);
julia> @benchmark logaddexp_old.(x, y)
BenchmarkTools.Trial: 2722 samples with 1 evaluation.
 Range (min … max):  1.813 ms …  2.113 ms  ┊ GC (min … max): 0.00% … 12.46%
 Time  (median):     1.827 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.834 ms ± 34.028 μs  ┊ GC (mean ± σ):  0.23% ±  1.54%

  ▁█▄▃▁                                                       
  ██████▆▅▄▃▃▂▂▂▂▂▂▂▁▁▁▁▂▁▁▁▁▁▂▁▂▂▁▁▁▁▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▂▂▂▂ ▃
  1.81 ms        Histogram: frequency by time        2.08 ms <

 Memory estimate: 781.36 KiB, allocs estimate: 4.
 julia> @benchmark logaddexp_new.(x, y)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  296.166 μs … 492.875 μs  ┊ GC (min … max): 0.00% … 39.48%
 Time  (median):     297.042 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   300.727 μs ±  24.536 μs  ┊ GC (mean ± σ):  1.07% ±  5.01%

  █▅                                                            ▁
  ██▇▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅▆▅█ █
  296 μs        Histogram: log(frequency) by time        486 μs <

 Memory estimate: 781.36 KiB, allocs estimate: 4.
```

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -309,9 +309,22 @@ Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and hand
 non-finite values.
 """
 function logaddexp(x::Real, y::Real)
-    # ensure Δ = 0 if x = y = ± Inf
-    Δ = x == y ? zero(x - y) : abs(x - y)
-    max(x, y) + log1pexp(-Δ)
+    if isnan(x) return x end
+    if isnan(y) return y end
+
+    diff = zero(promote_type(typeof(x), typeof(y)))
+    if x < y
+        diff = x - y
+        x = y
+    else
+        diff = y - x
+    end
+
+    if diff >= log(eps())
+        return x + log1p(exp(diff))
+    else
+        return x
+    end
 end
 
 Base.@deprecate logsumexp(x::Real, y::Real) logaddexp(x, y)


src/basicfuns.jl
```diff
@@ -309,9 +309,22 @@ Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and hand
 non-finite values.
 """
 function logaddexp(x::Real, y::Real)
-    # ensure Δ = 0 if x = y = ± Inf
-    Δ = x == y ? zero(x - y) : abs(x - y)
-    max(x, y) + log1pexp(-Δ)
+    if isnan(x) return x end
+    if isnan(y) return y end
```
devmotion, line 313: This causes type instabilities and can't be used unfortunately. It's bad that this is not covered by our tests, it seems.

src/basicfuns.jl
```diff
@@ -309,9 +309,22 @@ Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and hand
...
+
+    diff = zero(promote_type(typeof(x), typeof(y)))
+    if x < y
+        diff = x - y
+        x = y
+    else
+        diff = y - x
+    end
+
+    if diff >= log(eps())
```
devmotion, line 323: This is not correct since the check does not depend on the type of `diff` but the RHS is always a `Float64`. I am not happy about introducing any tolerances here in general, `iszero` is much cleaner.

src/basicfuns.jl
```diff
@@ -309,9 +309,22 @@ Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and hand
 non-finite values.
 """
 function logaddexp(x::Real, y::Real)
-    # ensure Δ = 0 if x = y = ± Inf
-    Δ = x == y ? zero(x - y) : abs(x - y)
-    max(x, y) + log1pexp(-Δ)
+    if isnan(x) return x end
+    if isnan(y) return y end
+
+    diff = zero(promote_type(typeof(x), typeof(y)))
```
devmotion, line 315: This is not needed since `diff` is defined below.

```suggestion
```

src/basicfuns.jl
```diff
@@ -309,9 +309,22 @@ Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and hand
...
+        diff = x - y
+        x = y
+    else
+        diff = y - x
+    end
+
+    if diff >= log(eps())
+        return x + log1p(exp(diff))
+    else
+        return x
```
devmotion, line 326: This causes a type instability as well for e.g. integers or rational numbers.

devmotion (changes requested):
Thanks for the PR. Unfortunately, it seems there are some problems with the proposed implementation which are not covered by our tests.

tpapp:
@devmotion:

> which are not covered by our tests

if you have the time, it would be great if you could add these tests

devmotion:
> if you have the time, it would be great if you could add these tests

I opened a PR: #42

src/basicfuns.jl
lucasondel: I understand that `diff < log(eps(typeof(diff)))` would be better,  however I don't see how one can use `iszero` in this context. Do you have a precise idea in mind ?

devmotion: Ah, yeah my bad I don't think `iszero` is helpful here. I am surprised that the branch is useful at all. It would be better to use `log1pexp` instead of `log1p(exp(...))`, as in the current implementation, since it is already optimized and avoids the `log1p` and `exp` calls if it is not required for numerical accuracy.

lucasondel:
This new version should be type-stable.

src/basicfuns.jl
```diff
@@ -309,9 +309,28 @@ Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and hand
 non-finite values.
 """
 function logaddexp(x::Real, y::Real)
-    # ensure Δ = 0 if x = y = ± Inf
-    Δ = x == y ? zero(x - y) : abs(x - y)
-    max(x, y) + log1pexp(-Δ)
+    T = float(Base.promote_typeof(x, y))
+    _logaddexp(T(x), T(y))
```
tpapp, line 313: you can just use something like
```julia
logaddexp(x, y) = logaddexp(promote(x, y)...)

function logaddexp(x::T, y::T) where {T <: AbstractFloat} 
   # float implementation follows here, can use T
end
```

lucasondel: Hmm, I think I need to use `logaddexp(float.(promote(x, y)...)` otherwise there will be some infinite recursion when using non-float values... But I get your point, I'll make the update.

lucasondel:
I kept the `Real` type specification in the global signature to be consistent with other functions.

src/basicfuns.jl
```diff
@@ -308,10 +308,25 @@ $(SIGNATURES)
 Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and handling
 non-finite values.
 """
-function logaddexp(x::Real, y::Real)
-    # ensure Δ = 0 if x = y = ± Inf
-    Δ = x == y ? zero(x - y) : abs(x - y)
-    max(x, y) + log1pexp(-Δ)
+logaddexp(x::Real, y::Real) = logaddexp(float.(promote(x, y))...)
```
devmotion, line 311: We can avoid the heavy broadcasting machinery here:

```suggestion
logaddexp(x::Real, y::Real) = logaddexp(map(float, promote(x, y))...)
```

src/basicfuns.jl
```diff
@@ -308,10 +308,25 @@ $(SIGNATURES)
...
+
+function logaddexp(x::T, y::T) where T <: AbstractFloat
```
devmotion, line 313: This is too strict, not every floating point number and output of `float` is of type `AbstractFloat`. Eg, usually dual numbers don't satisfy this property (typically they are subtypes of `Real` but not `AbstractFloat`).
Ie., this should be an internal method with signature

```suggestion
function _logaddexp(x::T, y::T) where {T<:Real}
```

src/basicfuns.jl
```diff
@@ -308,10 +308,25 @@ $(SIGNATURES)
...
+    if x < y
+        diff = x - y
+        maxval = y
+    elseif y <= x
+        diff = y - x
+        maxval = x
+    else
+        diff = zero(T)
+        maxval = T(NaN)
+    end
```
devmotion, line 323: This seems quite complicated. Is it helpful for improving performance?

src/basicfuns.jl
```diff
@@ -308,10 +308,25 @@ $(SIGNATURES)
...
+
+    if diff < log(eps(T)) || isnan(diff)
+        return maxval
+    else
+        return maxval + log1pexp(diff)
+    end
```
devmotion, line 329: This check is not correct, even for numbers smaller than `log(eps(T))` `log1pexp` returns non-zero values (see eg. https://github.com/JuliaStats/LogExpFunctions.jl/pull/43). With that PR it should be similarly performant (and more correct) to just use

```suggestion
    return maxval + log1pexp(diff)
```

devmotion (changes requested):
Can you update the benchmarks when you've updated the code?

src/basicfuns.jl
lucasondel: Yes, the point is to have a single branching instead of two when successively calling `max(x, y)` and `abs(x - y)`. Here is the benchmark when there is 2 branching (using `maxval = max(x, y)` and `diff = abs(x - y)`):
```
BenchmarkTools.Trial: 4680 samples with 1 evaluation.
 Range (min … max):  1.053 ms …  1.347 ms  ┊ GC (min … max): 0.00% … 20.78%
 Time  (median):     1.062 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.067 ms ± 28.219 μs  ┊ GC (mean ± σ):  0.34% ±  2.15%
```
and when there is only 1 branching:
```
BenchmarkTools.Trial: 8614 samples with 1 evaluation.
 Range (min … max):  567.500 μs … 827.459 μs  ┊ GC (min … max): 0.00% … 30.88% 
 Time  (median):     571.000 μs               ┊ GC (median):    0.00%            
 Time  (mean ± σ):   579.051 μs ±  33.001 μs  ┊ GC (mean ± σ):  0.70% ±  3.80%     
```

src/basicfuns.jl
lucasondel: I've merged locally with #43 but it doesn't help that much speed wise. In this settings the main bottleneck seems to be the evaluation of the threshold. 

When using `log(eps(T))`:
```
...
329     t = log(eps(T))                                                             
330     if diff < t || isnan(diff) #log(eps(T)) || isnan(diff)                      
331         return maxval                                                           
332     else                                                                        
333         return maxval + log1pexp(diff)                                          
334     end            
335 end                                                             
```
we get:
```
BenchmarkTools.Trial: 10000 samples with 1 evaluation.                                                                     
 Range (min … max):  333.208 μs … 566.667 μs  ┊ GC (min … max): 0.00% … 40.06%                                             
 Time  (median):     334.167 μs               ┊ GC (median):    0.00%                                                      
 Time  (mean ± σ):   338.132 μs ±  28.708 μs  ┊ GC (mean ± σ):  1.12% ±  5.14%
```

and when using the "correct" threshold:
```
...
328     logtwo = oftype(diff, IrrationalConstants.logtwo)                           
329     t = (exponent(nextfloat(zero(diff))) - 1) * logtwo                          
330     if diff < t || isnan(diff) #log(eps(T)) || isnan(diff)                      
331         return maxval                                                           
332     else                                                                        
333         return maxval + log1pexp(diff)                                          
334     end                                                                         
335 end
```
we get:
```
BenchmarkTools.Trial: 4710 samples with 1 evaluation.                                                                      
 Range (min … max):  1.050 ms …  1.360 ms  ┊ GC (min … max): 0.00% … 21.33%                                                
 Time  (median):     1.054 ms              ┊ GC (median):    0.00%                                                         
 Time  (mean ± σ):   1.060 ms ± 34.891 μs  ┊ GC (mean ± σ):  0.42% ±  2.58%
```

devmotion: The threshold is not evaluated in the calls but precomputed for common types such as `Float64`, `Float32`, and `Float16`. Even for other types the compiler should usually be able to precompute the thresholds and avoid computations in every call.

lucasondel: Yes, that was the point to use a predefined threshold... with the downside to lose a little bit in precision. For my particular use case (graph computation in the log-semiring), the precision loss is negligible compare to the significant gain. But this is probably too specific and it is better that the implementation in this package remains of general purpose and therefore favor precision. 

I'll simplify this part to: `return maxval + log1pexp(diff)`

lucasondel:
Here are the performance of the latest master branch:
```
julia> @benchmark logaddexp.(x, y)
BenchmarkTools.Trial: 3140 samples with 1 evaluation.
 Range (min … max):  1.557 ms …  2.169 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.575 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.590 ms ± 43.455 μs  ┊ GC (mean ± σ):  0.23% ± 1.55%
```

Here are the performance of this PR:
```
julia> @benchmark logaddexp.(x, y)
BenchmarkTools.Trial: 4539 samples with 1 evaluation.                                                                      
 Range (min … max):  1.077 ms …  1.451 ms  ┊ GC (min … max): 0.00% … 23.56%                                                
 Time  (median):     1.098 ms              ┊ GC (median):    0.00%                                                         
 Time  (mean ± σ):   1.100 ms ± 36.698 μs  ┊ GC (mean ± σ):  0.39% ±  2.43%                                                
```

NOTE: 
* for both cases: `x = 1000 * randn(10000)` and `y = 1000 * randn(10000)`
* I have run the test 3 times for each setting without significant difference in the results

src/basicfuns.jl
```diff
@@ -311,10 +311,20 @@ $(SIGNATURES)
...
+
+    if x < y
+        diff = x - y
+        maxval = y
+    else
+        diff = y - x
+        maxval = x
+    end
+
+    isnan(diff) ? maxval : maxval + log1pexp(diff)
```
devmotion, line 327: I'm not completely happy about the additional internal method, I don't think it should be needed here since we don't dispatch on it or reuse it in other function calls. Maybe this could be simplified to
```suggestion
function logaddexp(x::Real, y::Real)
    # Compute max = Base.max(x, y) and diff = x == y ? zero(x - y) : abs(x - y)
    # in a faster type-stable way
    a, b = promote(x, y)
    if a < b
        diff = b - a
        max = b
    else
        # ensure diff = 0 if a = b = ± Inf
        diff = a == b ? zero(a - b) : a - b
        max = !isnan(b) ? a : b
    end

    return max + log1pexp(diff)
```
?

In my benchmarks this leads to a similar improvement (BTW it's not necessary to use broadcasting in benchmarking and I also wonder if your benchmarks are off since you did not interpolate `x` and `y` but used global variables):
```julia
julia> @btime logaddexp(x, y) setup=(x = 1000 * randn(); y = 1000 * randn());
  3.686 ns (0 allocations: 0 bytes)

julia> @btime logaddexp_new(x, y) setup=(x = 1000 * randn(); y = 1000 * randn());
  2.612 ns (0 allocations: 0 bytes)

julia> @btime logaddexp.(x, y) setup=(x = 1000 * randn(10_000); y = 1000 * randn(10_000));
  125.475 μs (2 allocations: 78.17 KiB)

julia> @btime logaddexp_new.(x, y) setup=(x = 1000 * randn(10_000); y = 1000 * randn(10_000));
  82.540 μs (2 allocations: 78.17 KiB)
```

lucasondel: Yes, actually we can handle the type without resorting to yet another function. Some tests failed though with your proposed version so I kept the current "branching version" .

lucasondel:
@devmotion Is this version ok to be merged ?

src/basicfuns.jl
devmotion: > Some tests failed though with your proposed version so I kept the current "branching version" .

Seems I mistyped `diff` instead of `-diff` - then all tests pass :smile:

src/basicfuns.jl
```diff
@@ -312,9 +312,23 @@ Return `log(exp(x) + exp(y))`, avoiding intermediate overflow/undeflow, and hand
...
+    if a < b
+        diff = a - b
+        maxval = b
+    else
+        diff = b - a
+        maxval = a
+    end
+
+    # `diff` is NaN if both `a` and `b` are infinite
+    isnan(diff) ? maxval : maxval + log1pexp(diff)
```
devmotion, line 331: IMO the following is simpler and in particular avoids any explicit type calculations:
```suggestion
    # Compute max = Base.max(x, y) and diff = x == y ? zero(x - y) : -abs(x - y)
    # in a faster type-stable way
    a, b = promote(x, y)
    if a < b
        diff = a - b
        max = b
    else
        # ensure diff = 0 if a = b = ± Inf
        diff = a == b ? zero(a - b) : b - a
        max = !isnan(b) ? a : b
    end
    return max + log1pexp(diff)
```

On my computer it is also faster than the master branch and the current version of the PR:
```julia
julia> using LogExpFunctions, BenchmarkTools

julia> @btime logaddexp(x, y) setup=(x = 1000 * randn(); y = 1000 * randn()); # master branch
  2.370 ns (0 allocations: 0 bytes)                                    

julia> @btime logaddexp(x, y) setup=(x = 1000 * randn(); y = 1000 * randn()); # current proposal
  1.933 ns (0 allocations: 0 bytes)

julia> @btime logaddexp(x, y) setup=(x = 1000 * randn(); y = 1000 * randn()); # proposal
  1.504 ns (0 allocations: 0 bytes)  
```

devmotion (commented):
I fixed a typo in my previous suggestion, now all tests pass. I think it's a bit simpler, and on my computer it's even a bit faster (possibly due to fewer conditionals and `isnan` checks?).

lucasondel:
Hopefully, this one will be the final version :) !

devmotion (approved):
Looks good to me, thanks a lot!

PR #42, opened 2022-04-09
Check type stability and return type of `logaddexp`

devmotion:
This would have caught the issues I mentioned in https://github.com/JuliaStats/LogExpFunctions.jl/pull/41. I checked that the tests do not pass with the implementation in that PR.

### DIFF
--- a/test/basicfuns.jl
+++ b/test/basicfuns.jl
@@ -176,7 +176,8 @@ end
 end
 
 @testset "logsumexp" begin
-    for T1 in (Int, Float32, Float64), T2 in (Int, Float32, Float64)
+    Ts = (Int, Float32, Float64)
+    for T1 in Ts, T2 in Ts
         a = T1(2)
         b = T2(3)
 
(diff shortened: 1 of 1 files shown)


test/basicfuns.jl
```diff
@@ -176,8 +176,18 @@ end
 end
 
 @testset "logsumexp" begin
-    @test logaddexp(2.0, 3.0)     ≈ log(exp(2.0) + exp(3.0))
-    @test logaddexp(10002, 10003) ≈ 10000 + logaddexp(2.0, 3.0)
+    for T1 in (Int, Float32, Float64), T2 in (Int, Float32, Float64)
```
tpapp, line 179:
```suggestion
    Ts = (Int, Float32, Float64)
    for T1 in Ts, T2 in Ts
```
(in case we want to add types later)

devmotion:
Good to go @tpapp?

tpapp (approved):

tpapp:
yes, lgtm

PR #45, opened 2022-04-27
Add fallbacks for `log1pmx` and `logmxp1`

### LINKED ISSUE #44: `log1pmx` and `logmxp1` cannot be differentiated by `ForwardDiff`
```julia
using LogExpFunctions

using ForwardDiff
using Zygote
using Test

log1mexpm(x) = log1mexp(-x)
log2mexpm(x) = log2mexp(-x)

function test_ad_vs_zygote(AD)
    @testset verbose = true "LogExpFunctions" begin
        @testset "Single-argument functions" begin
            @testset "$f" for f in (                      
                log1pmx,
                logexpm1,
                logsumexp,
                xexpx,
                log1psq,
                logistic,
                invsoftplus,
                log2mexpm,
                logit,
                log1mexpm,
                logmxp1,
                xlogx,
                log1pexp,
                logcosh
            )
                for _ in 1:100
                    par = rand()
                    @test AD.derivative(f, par) ≈ only(only(Zygote.gradient(f ∘ only, [par])))
                end
            end
        end
        @testset "Two-argument functions" begin
            @testset "$f" for f in (                      
                xexpy,
                xlogy,
                xlog1py,
                logaddexp,
                logsubexp,
            )
                for _ in 1:100
                    par = rand(2)
                    @test AD.gradient(x -> f(x...), par) ≈ only(Zygote.gradient(x -> f(x...), par))
                end
            end
        end
        @testset "Vector-argument functions" begin
                @testset "$f" for f in (                      
                softmax,
            )
                for _ in 1:100
                    par = rand(3)
                    @test AD.jacobian(f, par) ≈ only(Zygote.jacobian(f, par))
                end
            end
        end
    end
    return nothing
end

test_ad_vs_zygote(ForwardDiff)
```
yields
```julia
Test Summary:               | Pass  Error  Total
LogExpFunctions             | 1800    200   2000
  Single-argument functions | 1200    200   1400
    log1pmx                 |         100    100
    logexpm1                |  100           100
    logsumexp               |  100           100
    xexpx                   |  100           100
    log1psq                 |  100           100
    logistic                |  100           100
    logexpm1                |  100           100
    log2mexpm               |  100           100
    logit                   |  100           100
    log1mexpm               |  100           100
    logmxp1                 |         100    100
    xlogx                   |  100           100
    log1pexp                |  100           100
    logcosh                 |  100           100
  Two-argument functions    |  500           500
  Vector-argument functions |  100           100
ERROR: Some tests did not pass: 1800 passed, 0 failed, 200 errored, 0 broken.
```

simsurace:
This looks like it would make sense, given that the other functions dispatch on `Real` argument, and it also fixes #44.

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -292,7 +292,15 @@ function logmxp1(x::Float64)
 end
 
 # Naive fallback
-logmxp1(x::Real) = (log(x) + one(x)) - x
+function logmxp1(x::Real)
+    one_x = one(x)
+    if 2 * x < one_x
+        # for small values of `x` the other branch returns non-finite values
+        return (log(x) + one_x) - x
+    else
+        return log1pmx(x - one_x)
+    end
+end
 
 # The kernel of log1pmx
 # Accuracy within ~2ulps for -0.227 < x < 0.315
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -246,6 +246,7 @@ $(SIGNATURES)
 Return `log(1 + x) - x`.
 
 Use naive calculation or range reduction outside kernel range.  Accurate ~2ulps for all `x`.
+This will fall back to the naive calculation for argument types different from `Float64`.
 """
 function log1pmx(x::Float64)
     if !(-0.7 < x < 0.9)
@@ -267,10 +268,14 @@ function log1pmx(x::Float64)
     end
 end
 
+# Naive fallback
+log1pmx(x::Real) = log1p(x) - x
+
 """
 $(SIGNATURES)
 
 Return `log(x) - x + 1` carefully evaluated.
+This will fall back to the naive calculation for argument types different from `Float64`.
 """
 function logmxp1(x::Float64)
     if x <= 0.3
@@ -285,6 +290,9 @@ function logmxp1(x::Float64)
         return log1pmx(x - 1.0)
     end
 end
+
+# Naive fallback
+logmxp1(x::Real) = (log(x) + one(x)) - x
 
 # The kernel of log1pmx
 # Accuracy within ~2ulps for -0.227 < x < 0.315
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LogExpFunctions"
 uuid = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 authors = ["StatsFun.jl contributors, Tamas K. Papp <tkpapp@gmail.com>"]
-version = "0.3.12"
+version = "0.3.13"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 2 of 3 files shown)


devmotion:
As I mentioned in #44, I think this PR is not the correct fix for #44, and hence the motivation is a bit different.

Regardless of that, I think it is reasonable to extend these functions to `::Real`. My only concern is that the proposed implementation might not be sufficiently accurate for `Float32` and `Float16` for which we provide optimized implementations for other functions. It would be good to check that it is sufficiently accurate for these types, and in particular the numerically problematic inputs.

simsurace:
I see. I need to look a bit into the floating-point issues here.
I might need help from someone more familiar with those things here. 
In particular, I don't really understand the optimized implementations for `Float64`.
I expect the problematic inputs to be at different locations for other floats.

simsurace:
Given that the PR https://github.com/JuliaDiff/DiffRules.jl/pull/82 is missing this fallback, I would propose to merge this and release a new version, open a separate issue for optimizations, and add those in a future PR.

devmotion:
We don't need the Float32 support in DiffRules, it's just a part of the tests that we can skip.

src/basicfuns.jl
```diff
@@ -286,6 +291,9 @@ function logmxp1(x::Float64)
     end
 end
 
+# Naive fallback
+logmxp1(x::Real) = (log(x) + one(x)) - x
```
devmotion, line 295: Should we use
```suggestion
logmxp1(x::Real) = log1pmx(x - one(x))
```
to take advantage of a (possibly specialized) implementation of `log1pmx`? That's also used in the branch for `x > 0.6` in the implementation for `Float64`.

simsurace: Right now, this fails for very small x:
```julia
julia> log1pmx(1e-100 - 1)
-Inf

julia> logmxp1(1e-100)
-229.25850929940458

julia> log(1e-100) + 1 - 1e-100
-229.25850929940458
```

simsurace: And for `Float32`:
```julia

julia> logmxp1(1f-40)
-91.10341f0

julia> log1pmx(1f-40 - 1)
-Inf32
```

devmotion: I guess that's why it's not used for small values for `Float64` :smile: 

Maybe we could use a very rough heuristic such as
```suggestion
function logmxp1(x::Real)
    one_x = one(x)
    if 2 * x < one_x
        # for small values of `x` the other branch returns non-finite values
        (log(x) + one_x) - x
    else
        log1pmx(x - one_x)
    end
end
```

devmotion (commented):
Seems fine in principle, I guess we can optimize the generic implementation in follow-up PRs if needed.

Can you add some tests with non-Float64 types (eg some Float32 values and compare their results with Float64 results) and update the version number?

devmotion (approved):
LGTM, thanks!

PR #48, opened 2022-05-02
sumlog

cscherrer:
This PR adds `sumlog`, a more efficient way to compute `sum(log, x)`. There's more discussion on this on Discourse here:
https://discourse.julialang.org/t/sum-of-logs/80370

EDIT:
I think we have a good enough understanding of what's possible to lay out some design criteria. That ought to be more efficient than taking each line of code in isolation.

As a starting point, I suggest
1. Whenever `sum(log ∘ f, x)` is defined, `sumlog(f, x)` should give the same result (within some tolerance, etc)
2. `sumlog(x) == sumlog(identity, x)`
3. `sumlog(f, x)` should support a `dims` keyword argument whenever `sum(log ∘ f, x)` does (i.e., when `x` is an AbstractArray)
4. `sumlog` should be type-stable and compiler-friendly when possible
5. `sumlog` should use the optimized method requiring a single `log` application, whenever that's possible.

@devmotion @mcabbott thoughts on these?


src/sumlog.jl
```diff
@@ -0,0 +1,36 @@
...
+
+"""
+$(SIGNATURES)
+
+Compute `sum(log.(X))`.
+
+`sum(log.(X))` can be evaluated much more quickly as `sum(log, X)`. However,
+this still requires computing `log` for each element of `X`. 
+
+`sumlog(X)` can be faster still, especially an the length of `X` increases.
```
devmotion, line 11: Maybe this could be shortened a bit to make it immediately clear what's the advantage of this function:
```suggestion
Compute `sum(log.(X))` with a single `log` evaluation.

This is faster than computing `sum(log.(X))` or even `sum(log, X)`, in particular as `X` increases.
```

src/sumlog.jl
```diff
@@ -0,0 +1,36 @@
...
+
+This works by representing the `j`th element of `X` as `xⱼ = aⱼ * 2 ^ bⱼ`,
```
devmotion, line 13:
```suggestion
This works by representing the `j`th element of `X` as ``x_j = a_j  2^b_j``,
```

src/sumlog.jl
```diff
@@ -0,0 +1,36 @@
...
+allowing us to write
+
+    ∑ⱼ log(xⱼ) = log(∏ⱼ aⱼ) + log(2) * ∑ⱼ bⱼ
+
```
devmotion, line 17:
````suggestion
```math
\\sum_j \\log{x_j} = \\log(\\prod_j a_j) + \\log{2} \\sum_j b_j
```
````

File: src/sumlog.jl [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+Since `log(2)` is constant, `sumlog` only requires a single `log` evaluation.
```
devmotion, line 18:
```suggestion
Since ``\\log{2}`` is constant, `sumlog` only requires a single `log` evaluation.
```

src/sumlog.jl
```diff
@@ -0,0 +1,36 @@
...
+"""
+function sumlog(x::AbstractArray{T}) where {T}
```
devmotion, line 20: It seems this will fail for non-floating point types `T` such as `Int`, etc., `BigFloat`, and for complex numbers?

src/sumlog.jl
```diff
@@ -0,0 +1,36 @@
...
+
+Since `log(2)` is constant, `sumlog` only requires a single `log` evaluation.
+"""
+function sumlog(x::AbstractArray{T}) where {T} 
+    sig = one(T) 
+    ex = zero(exponent(one(T)))
+    bound = floatmax(T) / 2 
+    for xj in x 
+        sig *= significand(xj)
+        ex += exponent(xj)
```
devmotion, line 26: Maybe one could use
```suggestion
function sumlog(x::AbstractArray{<:Real})
    T = float(eltype(x))
    sig = one(T) 
    ex = zero(exponent(sig))
    bound = floatmax(T) / 2 
    for xj in x
        float_xj = float(xj)
        sig *= significand(float_xj)
        ex += exponent(float_xj) 
```

test/sumlog.jl
```diff
@@ -0,0 +1,5 @@
+@testset "sumlog" begin
+    for x in [10 .* rand(1000), repeat([nextfloat(1.0)], 1000), repeat([prevfloat(2.0)], 1000)]
```
devmotion, line 2: Can you add more tests with other number types (`Int`, `Rational`, `BigFloat`,  etc.)?

devmotion (commented):
Nice, thanks!

I think there are some problems with the current implementation but I made some suggestions that hopefully can fix most of them.

Can you also update the version number and add it to the docs?

src/sumlog.jl
cscherrer: Personally, I tend to read docstrings from the REPL, where LaTeX doesn't render. But we can change it if you have a strong preference this way

src/sumlog.jl
cscherrer: Maybe even `<:AbstractFloat`? Then we could have a fall-back `sumlog(x) = sum(log, x)` so all cases should be covered

src/sumlog.jl
devmotion: In this package we use math-blocks and double ticks with LaTeX syntax for equations, so we should be consistent here. AFAIK this is also recommended in the Julia documentation and works better with Documenter.

src/sumlog.jl
devmotion: I thought about this but the main disadvantage would be that `Int`s, `Rational`s etc all would not benefit from the speedup even though their result is a floating point number.

cscherrer: Right, so having `sumlog(x) = sum(log, x)` would make them work, allowing the same code to work better across types. Maybe I'm misinderstanding you?

cscherrer:
@devmotion I guess we also need some ChainRules methods...

src/sumlog.jl
devmotion: I mean it would work but it would not be optimized. Even though I think there's no reason why it shouldn't, and rather it would be surprising that `sumlog(ones(10_000))` and `sumlog(ones(Int, 10_000))` would be have differently. With the suggestion above both would only use one log evaluation.

devmotion:
> @devmotion I guess we also need some ChainRules methods...

We could, but it's not necessary to do in this PR IMO.

Apart from the element type (see discussion above), I think the main problem left is that I assume the code is problematic for GPU arrays. Other array implementations in LohExpFunctions are written in a GPU-friendly way and should work with all array types.

test/sumlog.jl
cscherrer: added

src/sumlog.jl
cscherrer: I see. It looks like this does work better than the fallback `log(sum, x)`.

One concern I have with this is that `Real` is for symbolic types. I think I see a way to test for that without depending on them. I'll resolve this then try that so we can discuss separately

cscherrer:
> I think the main problem left is that I assume the code is problematic for GPU arrays. Other array implementations in LohExpFunctions are written in a GPU-friendly way and should work with all array types.

Do you see a nice way of doing this?

I see two other potential things to add:
1. Support for Tuples and NamedTuples
2. Support for calling `sumlog(f, x)`

Both less critical, so we can come back to them

cscherrer:
@devmotion one of these updates along the way killed all of the performance. Are you seeing this too? Need to backtrack a bit I guess, and maybe split the preproccessing into a separate function

src/sumlog.jl
```diff
@@ -0,0 +1,46 @@
...
+        if sig > bound
+            (a, b) = (significand(sig), exponent(sig))
+            sig = a
+            ex += b
+        end
+    end
+    log(sig) + logtwo * ex
+end
+
+# `T` might be a `Symbolics.Num`, which is not an `AbstractFloat`
```
devmotion, line 43: Possibly even more common would be dual numbers which are often parametric number types `<: Real` but not `<: AbstractFloat` with `float(dual)` also being a dual number (possibly with a different type parameter).

src/sumlog.jl
```diff
@@ -0,0 +1,46 @@
...
+@inline _sumlog(::Type{T}, x::AbstractArray{<:Real}) where {T} = sum(log, x)
+
+sumlog(x) = sum(log, x)
```
devmotion, line 46: This should not be needed, I think ? In LogExpFunctions most functions are restricted to `::Real` or `::AbstractArray{<:Real}` which is already covered above.

src/sumlog.jl
```diff
@@ -0,0 +1,46 @@
...
+        sig *= significand(float_xj)
+        ex += exponent(float_xj) 
+
+        # Significands are in the range [1,2), so multiplication will eventually overflow
+        if sig > bound
+            (a, b) = (significand(sig), exponent(sig))
+            sig = a
+            ex += b
+        end
+    end
```
devmotion, line 39: Can we maybe use `mapreduce` here? That should work on GPUs as well, similar to `logsumexp`.

```suggestion
    sig, ex = mapreduce(_sumlog_op; init=(one(T), zero(exponent(one(T))))) do xj
        float_xj = float(xj)
        return significand(float_xj), exponent(float_xj) 
    end
    return log(sig) + IrrationalConstants.logtwo * ex
end

function _sumlog_op((sig1, ex1), (sig2, ex2))
    sig = sig1 * sig2
    ex = ex1 + ex2
    # Significands are in the range [1,2), so multiplication will eventually overflow
    if sig > floatmax(typeof(sig)) / 2
        ex += exponent(sig)
        sig = significand(sig)
    end
    return sig, ex
end
```

src/sumlog.jl
```diff
@@ -0,0 +1,46 @@
+using IrrationalConstants: logtwo
+
```
devmotion, line 2: We already import the package, the convention here is to use explicit `IrrationalConstants.logtwo`.

devmotion (commented):
Maybe we can use `mapreduce` to support more general and in particular GPU arrays (similar to how we use `reduce` in `logsumexp`).

The function has to be added to the docs for the tests to pass.

cscherrer:
It got faster! Check it out:
```julia
julia> x = rand(1000);

julia> @btime sum(log, $x)
  6.362 μs (0 allocations: 0 bytes)
-1027.6

julia> @btime sumlog($x)
  986.857 ns (0 allocations: 0 bytes)
-1027.6
```

cscherrer:
> The function has to be added to the docs for the tests to pass.

I don't understand what you mean by this. Doscstrings are usually added automatically, what's left to do?

devmotion:
> What's left to do?

You have to add the function to docs/src/index.md.

test/sumlog.jl
```diff
@@ -0,0 +1,7 @@
+@testset "sumlog" begin
+    for T in [Int, Float16, Float32, Float64, BigFloat]
```
devmotion, line 2: I noticed that you removed the type restriction. Thus we should extend the tests and eg. check more general iterables (also with different types, abstract eltype etc since `sum(log, x)` would work for them) and also complex numbers.

cscherrer: `eltype` doesn't work well for `Base.Generator`s. Usually this is when I'd turn to something like
```julia
julia> Core.Compiler.return_type(gen.f, Tuple{eltype(gen.iter)})
Float64
```
We could instead have it fall back on the default, but I'd guess that will sacrifice performance.

mcabbott: I bet you could write an equally fast version which explicitly calls `iterate`, and widens if the type changes. (But usually the compiler will prove that it won't.)

One reason to keep `mapreduce` for arrays is that you can give it `dims`.

mcabbott: Should check more carefully, but this appears to work & is as fast as current version:
```julia
function sumlog(x)
    iter = iterate(x)
    if isnothing(iter)
        return eltype(x) <: Number ? zero(float(eltype(x))) : 0.0
    end
    x1 = float(iter[1])
    x1 isa AbstractFloat || return sum(log, x)
    sig, ex = significand(x1), exponent(x1)
    iter = iterate(x, iter[2])
    while iter !== nothing
        xj = float(iter[1])
        x1 isa AbstractFloat || return sum(log, x)  # maybe not ideal, re-starts iterator
        sig, ex = _sumlog_op((sig, ex), (significand(xj), exponent(xj)))
        iter = iterate(x, iter[2])
    end
    return log(sig) + IrrationalConstants.logtwo * ex
end

sumlog(f, x) = sumlog(Iterators.map(f, x))
sumlog(f, x, ys...) = sumlog(f(xy...) for xy in zip(x, ys...))
```

And for dims:
```julia
sumlog(x::AbstractArray{T}; dims=:) where T = _sumlog(float(T), dims, x)

function _sumlog(::Type{T}, ::Colon, x) where {T<:AbstractFloat}
    sig, ex = mapreduce(_sumlog_op, x; init=(one(T), zero(exponent(one(T))))) do xj
        float_xj = float(xj)
        significand(float_xj), exponent(float_xj) 
    end
    return log(sig) + IrrationalConstants.logtwo * ex
end

function _sumlog(::Type{T}, dims, x) where {T<:AbstractFloat}
    sig_ex = mapreduce(_sumlog_op, x; dims=dims, init=(one(T), zero(exponent(one(T))))) do xj
        float_xj = float(xj)
        significand(float_xj), exponent(float_xj) 
    end
    map(sig_ex) do (sig, ex)
        log(sig) + IrrationalConstants.logtwo * ex
    end
end
```
Should I make a PR to the PR?

test/sumlog.jl
```diff
@@ -0,0 +1,25 @@
+@testset "sumlog" begin
```
mcabbott, line 1: Some surprises:
```julia
julia> sumlog([1,2,-0.1,-0.2])
-3.2188758248682

julia> sumlog([1,2,NaN])
ERROR: DomainError with NaN:
Cannot be NaN or Inf.
Stacktrace:
  [1] (::Base.Math.var"#throw1#5")(x::Float64)
    @ Base.Math ./math.jl:845
  [2] exponent
    @ ./math.jl:848 [inlined]

julia> sumlog([-0.0])
ERROR: DomainError with -0.0:
Cannot be ±0.0.

julia> sum(log, -0.0)
-Inf
```

mcabbott: Calling `Base.Math._exponent_finite_nonzero` works around the NaN problem (although not for BigFloats).

Adding `xj < 0 && Base.Math.throw_complex_domainerror(:log, xj)` doesn't seem to cost much speed.

mcabbott:
For small arrays, calling `log(prod(x))` is a few times quicker again. 

On crude tests it doesn't seem less accurate when it works, although for large enough arrays it gives `Inf`. Can it be used as a fast path, maybe with a test for `Inf` afterwards to check if the "slow" path is required?

```julia
julia> for N in [10, 30, 100, 300, 1000]
         println()
         @show N
         xs = rand(N)
         @btime sum(log, $xs)
         @btime sumlog($xs)
         err = Float64(mean(abs(sum(log∘big, x) - sumlog(x)) for x in (rand(N) for _ in 1:100)))
         @show err
         @btime log(prod($xs))
         err2 = Float64(mean(abs(sum(log∘big, x) - log(prod(x))) for x in (rand(N) for _ in 1:100)))
         @show err2
       end

N = 10
  107.296 ns (0 allocations: 0 bytes)
  29.857 ns (0 allocations: 0 bytes)
err = 6.533083953902729e-16
  14.821 ns (0 allocations: 0 bytes)
err2 = 4.679379951051084e-16

N = 30
  241.358 ns (0 allocations: 0 bytes)
  70.983 ns (0 allocations: 0 bytes)
err = 2.2369534809836574e-15
  17.451 ns (0 allocations: 0 bytes)
err2 = 1.3853197780876373e-15

N = 100
  729.947 ns (0 allocations: 0 bytes)  # sum(log, x)
  218.037 ns (0 allocations: 0 bytes)  # sumlog(x), with a check for negative values (204.166 ns without)
err = 9.411386403503971e-15
  37.466 ns (0 allocations: 0 bytes)  # log(prod(x)), with no such check: 5x faster
err2 = 3.815901761052095e-15

N = 300
  2.213 μs (0 allocations: 0 bytes)
  623.547 ns (0 allocations: 0 bytes)
err = 2.239577037102115e-14
  108.782 ns (0 allocations: 0 bytes)
err2 = 1.4313227111152077e-14

N = 1000
  7.250 μs (0 allocations: 0 bytes)
  2.051 μs (0 allocations: 0 bytes)
err = 8.074581160990276e-14
  404.790 ns (0 allocations: 0 bytes)
err2 = Inf
```

devmotion:
> although for large enough arrays it gives Inf

I guess it should depend on the range of values in the array and could possibly happen even with very small arrays. That makes me wonder as well how much the errors in your example depends on the fact that you use only values in [0, 1).

mcabbott:
Sure. My guess is that never doing this for long vectors is probably a sensible heuristic, guessing they will often overflow (and are expensive to traverse twice). For short ones, perhaps it needs some smarter check to see if it's safe. There is some time available, e.g. plenty to check `sum(x)`, but small numbers will give problems too.

tpapp:
I am skeptical about including this functionality. I agree that in some contexts (eg likelihood / posterior density calculations) users  sum log densities, but usually they make a best effort to have those quantities *already* in logs. This is widely supported across the package ecosystem (cf `Distributions.logpdf`, etc). Taking a lot of `log`s explicitly in these calculations looks like a code smell.

@cscherrer, can you please explain where/why you need this? The discourse topic you link already takes that as a given.

cscherrer:
It's pretty common for log-density to include a `log(foo)` term, where `foo` might be `x`, `sigma`, or something else. Regression problems require a sum of these, so this problem comes up any time the `foo` is not constant.

devmotion:
> In GLM.jl, the gamma residual deviance is computed as

A bit unrelated, but using LogExpFunctions this could be implemented as
```julia
devresid(::Gamma, y::Real, μ::Real) = -2 * logmxp1(y / μ)
```
instead.

test/sumlog.jl
mcabbott: Also, note that tests right now only test Float64

test/sumlog.jl
mcabbott: https://github.com/cscherrer/LogExpFunctions.jl/pull/1 is a tidier version of the above.

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
 """
-$(SIGNATURES)
+    sumlog(X::AbstractArray{T}; dims)
```
devmotion, line 2:
```suggestion
    sumlog(x::AbstractArray{<:Real}; dims=:)
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
 
-Compute `sum(log.(X))` with a single `log` evaluation.
+Compute `sum(log.(X))` with a single `log` evaluation,
+provided `float(T) <: AbstractFloat`.
```
devmotion, line 5: This sounds as if the function requires `float(T) <: AbstractFloat` which is not the case.

```suggestion
Compute `sum(log, x; dims=dims)`.

If `float(eltype(x)) <: AbstractFloat` the computation is performed with a single `log` evaluation.
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
 
-This is faster than computing `sum(log.(X))` or even `sum(log, X)`, in
-particular as the size of `X` increases.
-
-This works by representing the `j`th element of `X` as ``x_j = a_j  2^{b_j}``,
+This is faster than computing `sum(log, X)`, especially for large `X`.
```
devmotion, line 7:
```suggestion
This is faster than computing `sum(log, x; dims=dims)`, especially for large `x`.
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+It works by representing the `j`th element of `X` as ``x_j = a_j  2^{b_j}``,
```
devmotion, line 8:
```suggestion
It works by representing the `j`th element of `x` as ``x_j = a_j  2^{b_j}`,
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
 ```math
 \\sum_j \\log{x_j} = \\log(\\prod_j a_j) + \\log{2} \\sum_j b_j
 ```
-Since ``\\log{2}`` is constant, `sumlog` only requires a single `log`
-evaluation.
 """
-function sumlog(x)
-    T = float(eltype(x))
-    _sumlog(T, values(x))
+sumlog(x::AbstractArray{T}; dims=:) where T = _sumlog(float(T), dims, x)
```
devmotion, line 14: I think it would be better to keep this restricted to `Real` as other methods in LogExpFunctions?

In any case, the type parameter is not necessary it seems and the dispatch on `dims` is not needed (see below).
```suggestion
sumlog(x::AbstractArray{<:Real}; dims=:) = _sumlog(float(eltype(x)), x; dims=dims)
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+
+function _sumlog(::Type{T}, ::Colon, x) where {T<:AbstractFloat}
+    sig, ex = mapreduce(_sumlog_op, x; init=(one(T), 0)) do xj
+        xj < 0 && Base.Math.throw_complex_domainerror(:log, xj)
+        float_xj = float(xj)
+        significand(float_xj), _exponent(float_xj) 
+    end
+    return log(sig) + IrrationalConstants.logtwo * T(ex)
 end
```
devmotion, line 23: I think this can be removed:
```suggestion
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
 
-@inline function _sumlog(::Type{T}, x) where {T<:AbstractFloat}
-    sig, ex = mapreduce(_sumlog_op, x; init=(one(T), zero(exponent(one(T))))) do xj
+function _sumlog(::Type{T}, dims, x) where {T<:AbstractFloat}
```
devmotion, line 25:
```suggestion
function _sumlog(::Type{T}, x; dims) where {T<:AbstractFloat}
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+    sig_ex = mapreduce(_sumlog_op, x; dims=dims, init=(one(T), 0)) do xj
+        xj < 0 && Base.Math.throw_complex_domainerror(:log, xj)
```
devmotion, line 27: I think it's better to not rely on some internal error function in base, it seems simple enough to throw a custom error message.
```suggestion
        xj < zero(xj) && throw(DomainError(xj, "log requires a non-negative argument"))
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
         float_xj = float(xj)
-        significand(float_xj), exponent(float_xj) 
+        significand(float_xj), _exponent(float_xj) 
+    end
+    map(sig_ex) do (sig, ex)
+        log(sig) + IrrationalConstants.logtwo * T(ex)
     end
```
devmotion, line 33:
```suggestion
    return log.(first.(sig_ex)) .+ IrrationalConstants.logtwo .* T.(last.(sig_ex))
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
-    return log(sig) + IrrationalConstants.logtwo * ex
 end
 
+# Fallback: `float(T)` is not always `<: AbstractFloat`, e.g. complex, dual numbers or symbolics
+_sumlog(::Type, dims, x) = sum(log, x; dims)
```
devmotion, line 37:
```suggestion
_sumlog(::Type, x; dims) = sum(log, x; dims=dims)
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+
 @inline function _sumlog_op((sig1, ex1), (sig2, ex2))
     sig = sig1 * sig2
+    # sig = ifelse(sig2<0, sig2, sig1 * sig2)
```
devmotion, line 41:
```suggestion
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
         sig = significand(sig)
     end
     return sig, ex
 end
 
-# `float(T)` is not always `isa AbstractFloat`, e.g. dual numbers or symbolics
-@inline _sumlog(::Type{T}, x) where {T} = sum(log, x)
+# The exported `exponent(x)` checks for `NaN` etc, this function doesn't, which is fine as `sig` keeps track.
+_exponent(x::Base.IEEEFloat) = Base.Math._exponent_finite_nonzero(x)
+Base.@assume_effects :nothrow _exponent(x::AbstractFloat) = Int(exponent(x))  # e.g. for BigFloat
```
devmotion, line 53: I'm strongly against using any internal non-exported functions in Base in a package such as LogExpFunctions. I really think we should just use `exponent`.

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+For any iterator which produces `AbstractFloat` elements,
+this can use `sumlog`'s fast reduction strategy.
+
+Signature with `f` is equivalent to `sum(log, map(f, x, ys...))`
+or `mapreduce(log∘f, +, x, ys...)`, without intermediate allocations.
+
+Does not accept a `dims` keyword.
+"""
+sumlog(f, x) = sumlog(Iterators.map(f, x))
+sumlog(f, x, ys...) = sumlog(f(xy...) for xy in zip(x, ys...))
```
devmotion, line 68: Is this really needed? Users could just create these iterators themselves and call `sumlog(x)`.

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+
+# Iterator version, uses the same `_sumlog_op`, should be the same speed.
+function sumlog(x)
+    iter = iterate(x)
+    if isnothing(iter)
```
devmotion, line 73: In older Julia versions
```suggestion
    if iter === nothing
```
should be better

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+        T = Base._return_type(first, Tuple{typeof(x)})
+        return T <: Number ? zero(float(T)) : 0.0
```
devmotion, line 75: Again, I don't think we should use such internals.
We could just return
```suggestion
        return zero(log(sum(x)))
```
or
```suggestion
        return zero(float(sum(x)))
```
Unfortunately, `sum(log, x)` doesn't seem to work for empty arrays - but this would even allow us to throw an error for empty arrays to be consistent with `sum(log, x)`.

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+    end
+    x1 = float(iter[1])
+    x1 isa AbstractFloat || return sum(log, x)
+    x1 < 0 && Base.Math.throw_complex_domainerror(:log, x1)
```
devmotion, line 79: Again, no internal function:
```suggestion
    x1 < zero(x1) && throw(DomainError(x1, "log requires a non-negative argument"))
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+    sig, ex = significand(x1), _exponent(x1)
+    nonfloat = zero(x1)
```
devmotion, line 81: Why is this needed?
```suggestion
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+    iter = iterate(x, iter[2])
+    while iter !== nothing
+        xj = float(iter[1])
+        if xj isa AbstractFloat
+            xj < 0 && Base.Math.throw_complex_domainerror(:log, xj)
+            sig, ex = _sumlog_op((sig, ex), (significand(xj), _exponent(xj)))
+        else
+            nonfloat += log(xj)
+        end
```
devmotion, line 90:
```suggestion
        xj isa AbstractFloat || return sum(log, x)
        xj < zero(xj) && throw(DomainError(xj, "log requires a non-negative argument"))
        sig, ex = _sumlog_op((sig, ex), (significand(xj), _exponent(xj)))
```

src/sumlog.jl
```diff
@@ -1,42 +1,94 @@
...
+        iter = iterate(x, iter[2])
+    end
+    return log(sig) + IrrationalConstants.logtwo * oftype(sig, ex) + nonfloat
```
devmotion, line 93:
```suggestion
    log_sig = log(sig)
    return log_sig + IrrationalConstants.logtwo * oftype(log_sig, ex)
```

src/sumlog.jl
mcabbott: Because there is no guarantee that half way through the iterator, you won't encounter one non-Float. There's a test for this exact case.

src/sumlog.jl
mcabbott: You already know sig is an AbstractFloat.

src/sumlog.jl
devmotion: Sure but in this case returning `sum(log, x)` seems sufficient?

src/sumlog.jl
mcabbott: This, BTW, was my alternative attempt at error checking. Instead of checking on every iteration, if you ensure that `sig` ends up negative, then the final `log` will throw the right error automatically. I wasn't able to make this as fast, which I'm surprised by, maybe it can be done.

I think maybe focusing on the big questions first might be better than immediately nitpicking.

The other big one is how to handle Float16. At the moment it's super-inaccurate. Maybe accumulation should happen in higher precision. Maybe that should happen for Float64 too.

src/sumlog.jl
devmotion: Yes, if we assume that `typeof(log(sig)) === typeof(sig)` this could be simplified to
```suggestion
    return log(sig) + IrrationalConstants.logtwo * oftype(sig, ex)
```
The assumption should hold in almost all cases, but it didn't seem harmful to not rely on this fact.

src/sumlog.jl
mcabbott: It's avoiding re-starting the iterator.

src/sumlog.jl
mcabbott: Real would be fine. The tests use complex numbers only as examples of things for which `float(x)` isn't an AbstractFloat. (Which in reality could be weird numbers like Dual, etc.)

src/sumlog.jl
mcabbott: Lets you use a do block, which is tidier than writing a generator yourself.

src/sumlog.jl
devmotion: But why is it not restarted in the case of the first element? In general, reiterating is not guaranteed to yield the same values.

Maybe we should just error if `float(xi)`  is not an AbstractFloat.

src/sumlog.jl
cscherrer: I agree with `x` instead of `X`. I tried to match the original docs, but now I see they use lowercase. I may have been looking at an old version.

I think in general it makes sense to match the Julia docs for `sum`. For arrays, that looks like `sum(f, A::AbstractArray; dims)`.

We've already discussed that we don't need `<:Real` for dispatch, and it disallows some types that would be useful to include.

cscherrer:
@tpapp just making sure you saw my response to your concern

tpapp:
@cssherrer: thanks for asking, and the original answer. I understand that this is sometimes done, but at the moment I am not convinced that the optimization is worth the extra code complexity. But if you have a user case, I am fine with including this.

Also, I don't know enough numerical analysis to understand the error behavior. More on that below.

tpapp:
I played around with random tests and the new implementation is sometimes better (**much** better) and sometimes worse than `sum(log, ...)`. It would be great to understand why and see if we can improve it further. Comparison of `log10(abs(difference to BigFloat))`:

plot

Code:

```julia
import Pkg; Pkg.activate(; temp = true)
Pkg.add(url = "https://github.com/cscherrer/LogExpFunctions.jl#master")
Pkg.add("PGFPlotsX")
using LogExpFunctions, PGFPlotsX, Random

function compare_randomly(rng, T, M = 2000)
    z = T.(randexp(M))
    precisely = sum(log, BigFloat.(z))
    precision(x) = T(log10(abs(x - precisely)))
    (precision(sum(log, z)), precision(sumlog(z)))
end

c16 = [compare_randomly(Random.GLOBAL_RNG, Float16) for _ in 1:1000]
c32 = [compare_randomly(Random.GLOBAL_RNG, Float32) for _ in 1:1000]
c64 = [compare_randomly(Random.GLOBAL_RNG, Float64) for _ in 1:1000]

@pgf Axis({ xlabel = "sum(log, ...)", ylabel = "sumlog(...)", only_marks,
            legend_pos = "outer north east", draw_opacity = 0.5, mark_size = 1},
          PlotInc(Coordinates(c16)),
          LegendEntry("Float16"),
          PlotInc(Coordinates(c32)),
          LegendEntry("Float32"),
          PlotInc(Coordinates(c64)),
          LegendEntry("Float64"))
```

cscherrer:
The exponent computation is always exact, so I'd guess the worst case for this matches the worst case for `prod`

mcabbott:
Maybe it's clearer to think in terms of how many `nextfloat`s it's wrong by. Here's some code to try, results I see:

* The errors are very strongly correlated with those of `sum(log, x)`. 
* Changing it to accumulate `sig::Float64` doesn't seem to help accuracy.
* When `log(prod(Float64, x))` is finite, it appears to be equally accurate.

So far just tried with with `scale * rand(N)`. With which, errors seem larger at N=10^3 than at 10^2 or 10^4, 10^5. Are there pathalogical choices of input which give much larger errors?
```julia
function countepsfrom(x::T, xtrue) where {T<:AbstractFloat}
    target = T(xtrue)
    for n in Iterators.flatten(zip(0:100, -1:-1:-100))
        nextfloat(x, n) === target && return n
    end
    nf = (target - x) / eps(x)
    return isfinite(nf) ? round(Int, nf) : nf
end
function countsumlog(xs::AbstractArray{T}, sig_bound...) where T
    ytrue = sum(log.(big.(xs)))
    ybase = sum(log, xs)
    yfast = sumlog(xs, sig_bound...)  # modified to allow sig = 1.0 of wider type, and to print when sig > bound
    yprod = T(log(prod(Float64, xs)))
    (base = countepsfrom(ybase, ytrue), fast = countepsfrom(ybase, yfast), prod = countepsfrom(ybase, yprod))
end
using Random

countsumlog([1,2,3.0])  # exact

[countsumlog(rand(1000)) for _ in 1:20]  # doesn't hit bound, worst error 14, very correlated

[countsumlog(rand(100)) for _ in 1:20]  # at this size, prod still works
[countsumlog(3200 * rand(100)) for _ in 1:20]  # at threshold, prod either Inf or good,

# Float32
countsumlog([1,2,3f0])  # exact

Random.seed!(1); [countsumlog(rand(Float32, 1000)) for _ in 1:20]  # all hit bound, larger errors, worst 17, perfectly correlated
Random.seed!(1); [countsumlog(rand(Float32, 1000), 1.0) for _ in 1:20] # accumulating in Float64 changes nothing

[countsumlog(rand(Float32, 100)) for _ in 1:20]  # here prod works, provided it accumulates Float64

# Float16
countsumlog(Float16[1,2,3])  # exact

Random.seed!(2); [countsumlog(rand(Float16, 1000)) for _ in 1:20]  # ditto, worst -18

[countsumlog(rand(Float16, 100)) for _ in 1:20]  # prod works fine
[countsumlog(3200 * rand(Float16, 100)) for _ in 1:20]  # same threshold for prod, since it uses Float64
```

mcabbott:
The answer for pathalogical cases is numbers near to 1, one of which was in the tests. With my code just above:
```julia
julia> x1p = fill(nextfloat(1.0), 100);

julia> countsumlog(x1p)
(base = 0, fast = -77, prod = -77)

julia> x1m = fill(prevfloat(1.0), 100);

julia> countsumlog(x1m)
(base = 0, fast = -1970324836974592, prod = -39)

julia> (sum(log, x1m), sumlog(x1m))
(-1.1102230246251565e-14, -1.4210854715202004e-14)
```

ps. I think @tpapp's code needs `Pkg.add(url = "https://github.com/cscherrer/LogExpFunctions.jl", rev="master")` to run, and probably wants `z = T.(randexp(rng, M))` else all the RNG stuff does nothing.

tpapp:
@mcabbott: thanks for the correction, indeed I didn't use the RNG.

I think you are right. You can make a pathological example with eg
```julia
z = [1 + (eps(T) * ((i % 2) - 1)) for i in 1:1000]
```
for `T = Float64` etc, where `sum(log, ...)` does dramatically better. My intuition for this is that all the precision is at the end of the mantissa, and multiplication does not preserve enough. I suspect that anything close to powers of 2 would do badly.

cscherrer:
@devmotion @mcabbott @tpapp did you see the updated description in the OP? I think we could make some more progress by stepping back and reaching agreement on this. Then hopefully any questions on the code will be more stylistic. It's more like Test-Driven Development - we could even take an explicit TDD approach if you like.

BTW, I've thought of another example. The determinant of a matrix is the product of its eigenvalues, so the log-determinant is the sum of its log-eigenvalues. This example is interesting, because in this case we really want the "unsafe" version allowing negative eigenvalues. If our matrix happens to be triangular, we get a very fast way to compute `logdet`.

MeasureTheory includes parameterization according to the affine transform, so for example `x = σ * z + μ` where `z` is iid normal. In this case, `σ` is often triangular, so we get nice speedups.

mcabbott:
The goals seem fine to me.
* Does goal 1 imply that `0.0` and `NaN` etc. should propagate as usual? They give errors with `exponent`, which I tried to work around, but it won't work for e.g. high-precision AbstractFloats in various packages. Not sure I timed a version with an `if` to shield this. 
* I don't see how to do `sum(log ∘ f, x; dims)` without type-inference, since `f` could be (say) real-to-complex. Although possibly you could hide this my making it `sumlog!(f, y, x)`, constructing `y` by the same path that `mapreduce` uses, and dispatching on `y::AbstractArray{<:AbstractFloat}`.

devmotion:
I think it might be easier to focus on goals 1 (accuracy and consistency with `sum(log, x)`), 4 (type stability), 5 (performance), and GPU compatibility first. Supporting `sumlog(f, x)` and optional `dims` arguments seems less relevant initially.

Generally, I can see that the function can be useful in some cases but I would like to avoid that code complexity is increased too much in this package, so I think a simple implementation should be another main goal. IMO the code for `logsumexp` is already quite complex and hence difficult to maintain but probably this is justified by the popularity of this particular function.

cscherrer:
> * Does goal 1 imply that `0.0` and `NaN` etc. should propagate as usual?

Ideally, yes. But none of these are requirements in any way. The idea is more that if we start with an idealized wish list, it might be easier to talk about the design space and decide together where to make compromises.

Maybe this is just me, but after ten or so updates I find it too easy to get lost in the weeds. Maybe this can help keep us form going in circles in the discussion.

> * I don't see how to do `sum(log ∘ f, x; dims)` without type-inference

I think to start we should focus on the real case. We can come back to complex numbers - maybe this could be a kwarg or optional type parameter, or even a separate function. 

> Supporting `sumlog(f, x)` and optional `dims` arguments seems less relevant initially.

If we use `mapreduce`, I think we get `dims` support almost for free, is that right? 

> Generally, I can see that the function can be useful in some cases but I would like to avoid that code complexity is increased too much in this package, so I think a simple implementation should be another main goal.

I like "as simple as possible, but no simpler". I can understand wanting to avoid `Base._return_type`, and to lean toward higher-order functions to help with AD. But one concern with simplicity is the potential for others to re-implement to avoid any shortcomings. IMO some degree of complexity is better than simple code no one uses.

------

Also... We could consider changing this function name to `logprod`. Note that `sum(log, x) ≈ log(prod(x))`, with `sum(log, x)` having a more restricted domain (no negative reals allowed), and `log(prod, x)` being faster, but much more likely to overflow or underflow.

So if it's easy to set it up so "double negatives cancel", `logprod` might be a better name.

cscherrer:
This also points to another application - maybe you really want to just compute a product, but you'd like to avoid underflow and overflow. So for example @cjdoris's [LogarithmicNumbers.jl](https://github.com/cjdoris/LogarithmicNumbers.jl) would seem to benefit from adding
```julia
Base.prod(::Type{ULogarithmic}, x) = exp(ULogarithmic, logprod(x))
```

devmotion:
> avoid Base._return_type

Yes, this is part of the code complexity goal but will also improve stability of LogExpFunctions. All such internal functions and "hacks" should be removed from the PR, in particular since it seems they can be avoided easily. Even standard libraries such as Statistics don't use _return_type to handle empty iterators, see eg https://github.com/JuliaLang/Statistics.jl/blob/cdd95fea3ce7bf31c68e01412548688fbd505903/src/Statistics.jl#L204 and https://github.com/JuliaLang/Statistics.jl/blob/cdd95fea3ce7bf31c68e01412548688fbd505903/src/Statistics.jl#L170.

tpapp:
@cscherrer: Regarding stepping back and agreement: I always think in terms of costs and benefits (code complexity and maintainability vs how useful the code is), and personally I would just go to logs as soon as possible, even at a slight performance cost. But if you really need this, I am fine with including it.

Regarding the goals: 

1. I would rename to `logprod`, conveys the underlying algorithm and precision trade-offs better,

2. I think that `dims` and `foo(f, x)` are unnecessary, and it is silly that each function replicates the boilerplate for this, given that Julia has much nicer mechanisms now for these, but I understand that whenever we leave these out someone will complain

3. I would be happy with a robust and reasonably accurate `logprod` that is approximately `sum(log, ...)`, with the understanding that someones one is better than the other. All algorithms have trade-offs and that's fine. Maybe we should document them though.

cscherrer:
I've pushed another version. This time
1. It's `logprod` instead of `sumlog`
2. `sumlog` is still there for now, for easy comparison
3. I remembered about `frexp`. @mcabbott this behaves better for subnormals
4. I dropped `dims`, etc. I think we have a better understanding now of each other's priorities. I'm still in favor of more functionality, but we can start simple and get tests etc going. That will make it easier to weigh any drawbacks of adding more functinoality.
5. I think this is a good candidate to go in Base. If we make it `logabsprod` it can be a big help to speed up `logabsdet`:

```julia
julia> x = LowerTriangular(randn(1000,1000));

julia> using LinearAlgebra

julia> using BenchmarkTools

julia> @btime logabsdet($x)
  8.687 μs (0 allocations: 0 bytes)
(-631.836, -1.0)

julia> d = diag(x);

julia> @btime logabsprod($d)
  1.202 μs (0 allocations: 0 bytes)
(-631.836, 1.0)
```
I cheated here a little, since we don't have (that I know of) a lazy `diag` in Base.

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+For any iterator which produces `AbstractFloat` elements,
+this can use `logprod`'s fast reduction strategy.
+
+Signature with `f` is equivalent to `sum(log, map(f, x, ys...))`
+or `mapreduce(log∘f, +, x, ys...)`, without intermediate allocations.
+
+Does not accept a `dims` keyword.
+"""
+logprod(f, x) = logprod(Iterators.map(f, x))
+logprod(f, x, ys...) = logprod(f(xy...) for xy in zip(x, ys...))
```
devmotion, line 53:
```suggestion
```

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+
+# Iterator version, uses the same `_logprod_op`, should be the same speed.
+function logprod(x)
```
devmotion, line 56:
```suggestion
# Iterator version, uses the same `_logabsprod_op`, should be the same speed.
function logabsprod(x)
```

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+    iter = iterate(x)
+    if isnothing(iter)
```
devmotion, line 58:
```suggestion
    if iter === nothing
```

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+        T = Base._return_type(first, Tuple{typeof(x)})
+        return T <: Number ? zero(float(T)) : 0.0
```
devmotion, line 60: No internals, please. `prod` works for empty iterators and arrays, so we could just use
```suggestion
        y = prod(x)
        return log(abs(y)), sign(y)
```

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+    end
+    x1 = float(iter[1])
+    x1 isa AbstractFloat || return sum(log, x)
```
devmotion, line 63:
```suggestion
    if !(x1 isa AbstractFloat)
        y = prod(x)
        return log(abs(y)), sign(y)
    end
```

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+    x1 < 0 && Base.Math.throw_complex_domainerror(:log, x1)
```
devmotion, line 64: No internals, please.
```suggestion
```

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+    sig, ex = significand(x1), _exponent(x1)
```
devmotion, line 65:
```suggestion
    sig, ex = frexp(x1)
```

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+    nonfloat = zero(x1)
```
devmotion, line 66:
```suggestion
```

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+    iter = iterate(x, iter[2])
+    while iter !== nothing
+        xj = float(iter[1])
+        if xj isa AbstractFloat
+            sig, ex = _logprod_op((sig, ex), (significand(xj), _exponent(xj)))
+        else
+            nonfloat += log(xj)
```
devmotion, line 73:
```suggestion
            y = prod(x)
            return log(abs(y)), sign(y)
```

cscherrer:
> `logprod` is a neat idea to avoid checks.

Thanks! I think the name is more natural too :) 

> `frexp` is also clearly what we were looking for. Does this have any effect on speed?

I had assumed if anything it might be more efficient, but maybe not. Here's a quick check:
```julia
julia> @benchmark frexp(x) setup=(x=100 * rand())
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.063 ns (0.00% GC)
  median time:      2.084 ns (0.00% GC)
  mean time:        2.085 ns (0.00% GC)
  maximum time:     6.112 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark (significand(x), exponent(x)) setup=(x=100 * rand())
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.613 ns (0.00% GC)
  median time:      1.633 ns (0.00% GC)
  mean time:        1.640 ns (0.00% GC)
  maximum time:     10.330 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```

May not be a real effect since this is sub-nanosecond.

> I am lost in all the noise on minor details, but checking `sig > floatmax(typeof(sig)) / 2` is now the wrong thing, as it will overflow towards zero.

I had missed this, but @devmotion caught it too. Very surprising it would be so different. 

I'm also wondering... Was this wrong in the first place? A factor of 2 should work if we assume it's sequential. But some implementations might exploit the associativity. In this case I think we need `sqrt`, or better, to look at `exponent(sig)`. But with that we end up in the renormalization branch more often, so we're only about twice as fast as `sum(log, x)`

> `logprod` should really have a case which tries `log(prod())` first on small enough arrays, as this is much faster. (And it should advertise itself as being less prone to overflow than `log(prod`, rather than as being faster.)

Both great points.

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
+"""
+    logprod(X::AbstractArray{T}; dims)
+
+Compute `sum(log.(X))` with a single `log` evaluation,
+provided `float(T) <: AbstractFloat`.
+
+This is faster than computing `sum(log, X)`, especially for large `X`.
```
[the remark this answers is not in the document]
cscherrer: (outdated)

src/logprod.jl
```diff
@@ -0,0 +1,78 @@
...
+
+export logabsprod
+
+function logabsprod(x::AbstractArray{T}) where {T}
+    sig, ex = mapreduce(_logabsprod_op, x; init=frexp(one(T))) do xj
+        float_xj = float(xj)
+        frexp(float_xj)
+    end
+    sgn = signbit(sig) ? -one(T) : one(T)
+    return (log(abs(sig)) + IrrationalConstants.logtwo * T(ex), sgn)
```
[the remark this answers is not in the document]
cscherrer: Ok, I hadn't checked this before but it's nice to know
```julia
julia> typeof(sign(one(Float16)))
Float16
```

src/logprod.jl
cscherrer: @devmotion can you help me understand why you prefer this?

src/logprod.jl
cscherrer: Agreed, this was left over - I didn't mess with the iterator version much yet

src/logprod.jl
devmotion: Because it's more efficient in older Julia versions (IIRC it doesn't matter in more recent versions, probably >= 1.6?).

devmotion: See e.g. https://github.com/JuliaLang/julia/issues/35585

oscardssmith:
do we want to merge this?

cscherrer:
Same. I expected this would be relatively simple, and was surprised by the number of cases to worry about. I agree merging and then handling various cases as they come up is reasonable. It looks like the tests need some update though, they're currently failing because they use `sumlog` (now undefined) instead of the updated name `logprod`.

devmotion:
I don't remember the details either. It seems there are unaddressed comments and the tests don't pass yet, so I guess the PR needs some updates and another round of review before merging.

tpapp:
Friendly ping: this is a great contribution and it would be unfortunate to leave it dormant. @cscherrer, if you have the time to address the May 10 review comments by @devmotion and fix CI, I would be happy to merge.

tpapp:
Closing because of no activity.

That said, this is useful functionality, so if there is in interest in reviving the feature, just please open a new PR.

PR #54, opened 2022-07-07
Add `iszero(x)` branches to `xlogy` and `xlog1py`

simsurace:
Some of the tests don't pass yet. Maybe a finite differencing setup issue, I don't know. I'll take another look tomorrow or next week.

### DIFF
--- a/src/chainrules.jl
+++ b/src/chainrules.jl
@@ -19,9 +19,15 @@ end
 function _Ω_∂_xlogy(x::Real, y::Real)
     logy = log(y)
     z = x * logy
-    Ω = iszero(x) && !isnan(y) ? zero(z) : z
-    ∂x = logy
-    ∂y = x / y
+    w = x / y
+    if iszero(x) && !isnan(y)
+        Ω = zero(z)
+        ∂y = zero(w)
+    else
+        Ω = z
+        ∂y = iszero(y) ? oftype(w, NaN) : w
+    end
+    ∂x = iszero(y) ? oftype(logy, NaN) : logy
     return Ω, ∂x, ∂y
 end
 function ChainRulesCore.frule((_, Δx, Δy), ::typeof(xlogy), x::Real, y::Real)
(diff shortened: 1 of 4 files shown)


src/chainrules.jl
```diff
@@ -19,9 +19,15 @@ end
 function _Ω_∂_xlogy(x::Real, y::Real)
     logy = log(y)
     z = x * logy
-    Ω = iszero(x) && !isnan(y) ? zero(z) : z
-    ∂x = logy
-    ∂y = x / y
+    w = x / y
+    if iszero(x) && !isnan(y)
+        Ω = zero(z)
+        ∂y = zero(w)
```
devmotion, line 25: This is what I had in mind and think should be added. I hadn't managed to open a PR since I'm on vacation until beginning of August and only have limited time right now 🙂

src/chainrules.jl
```diff
@@ -19,9 +19,15 @@ end
...
+    else
+        Ω = z
+        ∂y = iszero(y) ? oftype(w, NaN) : w
+    end
+    ∂x = iszero(y) ? oftype(logy, NaN) : logy
```
devmotion, line 30: I'm not sure if these should be added, even if it makes the rule match finite differencing. NaN is usually not very useful as it infects everything.  So eg the forward rule will always yield NaN for any of these conditions. See eg https://juliadiff.org/ChainRulesCore.jl/stable/maths/nondiff_points.html.

src/chainrules.jl
simsurace: Great!

src/chainrules.jl
simsurace: This is a bit mysterious ti me. On the (seemingly equivalent) branches in the rrule for logpdf of NegativeBinomial we wanted NaN. Was this something specific to Distributions.jl?

simsurace: I'm happy to change this but I'm unsure about the discrepancy to Distributions.jl

PR #56, opened 2022-07-27
Add `log(-log(1 - x))` and `1 - exp(-exp(x))`

ararslan:
Still needs tests and a better implementation of the derivative for `cexpexp`. The implementations of the functions themselves come from GLM.

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -407,3 +407,17 @@ function _softmax!(r, x, dims)
     end
     return r
 end
+
+"""
+$(SIGNATURES)
+
+Compute the complementary log-log, `log(-log(1 - x))`.
+"""
+cloglog(x) = log(-log1p(-x))
+
+"""
+$(SIGNATURES)
+
+Compute the complementary double exponential, `1 - exp(-exp(x))`.
+"""
+cexpexp(x) = -expm1(-exp(x))
--- a/test/basicfuns.jl
+++ b/test/basicfuns.jl
@@ -412,13 +412,25 @@ end
 end
 
 @testset "cloglog and cexpexp" begin
+    cloglog_big(x::T) where {T} = T(log(-log(1 - BigFloat(x))))
+    cexpexp_big(x::T) where {T} = 1 - exp(-exp(BigFloat(x)))
+
+    for x in 0.1:0.1:0.9
+        @test cloglog(x) ≈ cloglog_big(x)
+        @test cexpexp(x) ≈ cexpexp_big(x)
+    end
+    for _ in 1:10
+        randf = rand(Float64)
+        @test cloglog(randf) ≈ cloglog_big(randf)
+        randi = rand(Int)
+        @test cexpexp(randi) ≈ cexpexp_big(randi)
+    end
+
     @test cloglog(0) == -Inf
     @test cloglog(1) == Inf
     @test cloglog((ℯ - 1) / ℯ) == 0
-    @test cloglog(0.5) == -0.366512920581664327012439158232669469454263447837105263053
 
     @test cexpexp(Inf) == 1.0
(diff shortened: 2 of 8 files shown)


src/basicfuns.jl
```diff
@@ -407,3 +407,17 @@ function _softmax!(r, x, dims)
     end
     return r
 end
+
+"""
+$(SIGNATURES)
+
+Compute the complementary log-log, `log(-log(x))`.
```
tpapp, line 414:
```suggestion
Compute the complementary log-log, `log(-log(1-x))`.
```

tpapp (commented):
Just a minor correction, let us know when it is ready for a full review.

src/basicfuns.jl
```diff
@@ -407,3 +407,17 @@ function _softmax!(r, x, dims)
...
+"""
+cloglog(x) = log(-log1p(-x))
+
+"""
+$(SIGNATURES)
+
+Compute the complementary double exponential, `exp(-exp(x))`.
```
devmotion, line 421: Same here, it seems this should be

```suggestion
Compute the complementary double exponential, `1 - exp(-exp(x))`.
```

src/basicfuns.jl
ararslan: Wow, this is why I shouldn't make PRs after 11pm. I was thinking of the complementary log-log in terms of its use in survival analysis, which is `log(-log(x))`, but copied the implementation for its use as a link function, which is indeed `log(-log(1 - x))`. Would it be worthwhile to have both, do you think?

ararslan:
Should be ready for a full review now!

src/basicfuns.jl
tpapp: Generally, I think one should add things which have a use case. If you need both, sure, just add it, otherwise it can wait for another PR.

test/basicfuns.jl
```diff
@@ -410,3 +410,15 @@ end
     @test axes(s, 1) == OffsetArrays.IdOffsetRange(-2:0)
     @test collect(s) ≈ softmax(1:3)
 end
+
+@testset "cloglog and cexpexp" begin
+    @test cloglog(0) == -Inf
+    @test cloglog(1) == Inf
+    @test cloglog((ℯ - 1) / ℯ) == 0
+    @test cloglog(0.5) == -0.366512920581664327012439158232669469454263447837105263053
```
tpapp, line 418: Instead of harcoding constants like the last one, perhaps it would be better to compare to something like
```julia
cloglog_big(x::T) where T = T(log(-log(1 -  BigFloat(x))))
```
with `≈`, using `0.1, 0.2, ..., 0.9` and a bunch of random numbers.

test/basicfuns.jl
```diff
@@ -410,3 +410,15 @@ end
...
+
+    @test cexpexp(Inf) == 1.0
+    @test cexpexp(-Inf) == 0.0
+    @test cexpexp(0) == (ℯ - 1) / ℯ
+    @test cexpexp(-0.366512920581664327012439158232669469454263447837105263053) == 0.5
```
tpapp, line 423: Same here, except that it could be done in the same loop as the above, basically testing
```julia
y = cloglog(x)
@test y ≈ cloglog_big(x)
@test cexpexp(y) ≈ x
```
with suitable tolerances.

tpapp (changes requested):
LGTM, made some testing suggestions.

src/basicfuns.jl
ararslan: My uses for both would be in Survival.jl and GLM.jl but might as well stick with what's implemented in this PR for now and do a follow-up if I can't just rearrange some stuff in Survival to use this formulation.

tpapp (approved):
LGTM, thanks!

PR #57, opened 2022-08-19
Add iszero(x) branches to xlogy and xlog1py

simsurace:
Apologies for the new PR, but I accidentally deleted my remote branch. This replaces #54

### DIFF
--- a/src/chainrules.jl
+++ b/src/chainrules.jl
@@ -57,11 +57,7 @@ function _Ω_∂_xlog1py(x::Real, y::Real)
 end
 function ChainRulesCore.frule((_, Δx, Δy), ::typeof(xlog1py), x::Real, y::Real)
     Ω, ∂x, ∂y = _Ω_∂_xlog1py(x, y)
-    if iszero(Δx)
-        ΔΩ = ∂y * Δy
-    else
-        ΔΩ = muladd(∂x, Δx, ∂y * Δy)
-    end
+    ΔΩ = muladd(∂x, Δx, ∂y * Δy)
     return Ω, ΔΩ
 end
 function ChainRulesCore.rrule(::typeof(xlog1py), x::Real, y::Real)
--- a/src/chainrules.jl
+++ b/src/chainrules.jl
@@ -32,11 +32,7 @@ function _Ω_∂_xlogy(x::Real, y::Real)
 end
 function ChainRulesCore.frule((_, Δx, Δy), ::typeof(xlogy), x::Real, y::Real)
     Ω, ∂x, ∂y = _Ω_∂_xlogy(x, y)
-    if iszero(Δx)
-        ΔΩ = ∂y * Δy
-    else
-        ΔΩ = muladd(∂x, Δx, ∂y * Δy)
-    end
+    ΔΩ = muladd(∂x, Δx, ∂y * Δy)
     return Ω, ΔΩ
 end
 function ChainRulesCore.rrule(::typeof(xlogy), x::Real, y::Real)
@@ -61,7 +57,11 @@ function _Ω_∂_xlog1py(x::Real, y::Real)
 end
 function ChainRulesCore.frule((_, Δx, Δy), ::typeof(xlog1py), x::Real, y::Real)
     Ω, ∂x, ∂y = _Ω_∂_xlog1py(x, y)
-    ΔΩ = muladd(∂x, Δx, ∂y * Δy)
+    if iszero(Δx)
+        ΔΩ = ∂y * Δy
+    else
+        ΔΩ = muladd(∂x, Δx, ∂y * Δy)
+    end
     return Ω, ΔΩ
 end
 function ChainRulesCore.rrule(::typeof(xlog1py), x::Real, y::Real)
--- a/src/chainrules.jl
+++ b/src/chainrules.jl
@@ -44,8 +44,7 @@ end
 function _Ω_∂_xlog1py(x::Real, y::Real)
     log1py = log1p(y)
     z = x * log1py
-    yp1 = 1 + y
-    w = x / yp1
+    w = x / (1 + y)
     if iszero(x) && !isnan(y)
         Ω = zero(z)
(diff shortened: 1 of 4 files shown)


src/chainrules.jl
```diff
@@ -38,9 +44,16 @@ end
 function _Ω_∂_xlog1py(x::Real, y::Real)
     log1py = log1p(y)
     z = x * log1py
-    Ω = iszero(x) && !isnan(y) ? zero(z) : z
+    yp1 = 1 + y
+    w = x / yp1
```
devmotion, line 48: It seems `yp1` is not used anywhere else:
```suggestion
    w = x / (1 + y)
```

test/chainrules.jl
```diff
@@ -2,6 +2,12 @@
     x = rand()
     test_frule(xlogx, x)
     test_rrule(xlogx, x)
+
```
devmotion, line 5:
```suggestion

# Test `iszero(x)` branches
```

test/chainrules.jl
```diff
@@ -2,6 +2,12 @@
...
+    test_frule(xlogy, 0.0, 1.0; fdm = forward_fdm(5, 1), nans = true)
+    test_rrule(xlogy, 0.0, 1.0; fdm = forward_fdm(5, 1), nans = true)
+
+    test_frule(xlog1py, 0.0, -1.0; fdm = forward_fdm(5, 1), nans = true)
+    test_rrule(xlog1py, 0.0, -1.0; fdm = forward_fdm(5, 1), nans = true)
```
devmotion, line 10:
```suggestion
    test_rrule(xlog1py, 0.0, -1.0; fdm = forward_fdm(5, 1), nans = true)

```

test/chainrules.jl
```diff
@@ -2,6 +2,12 @@
     x = rand()
     test_frule(xlogx, x)
     test_rrule(xlogx, x)
+
+    test_frule(xlogy, 0.0, 1.0; fdm = forward_fdm(5, 1), nans = true)
+    test_rrule(xlogy, 0.0, 1.0; fdm = forward_fdm(5, 1), nans = true)
+
```
devmotion, line 8:
```suggestion
```

devmotion (commented):
Hmm it seems there are still some test failures.

simsurace:
Hmm I saw them pass locally. Will get back to this...

src/chainrules.jl
```diff
@@ -38,9 +44,15 @@ end
 function _Ω_∂_xlog1py(x::Real, y::Real)
     log1py = log1p(y)
     z = x * log1py
-    Ω = iszero(x) && !isnan(y) ? zero(z) : z
+    w = x / (1 + y)
+    if iszero(x) && !isnan(y)
+        Ω = zero(z)
+        ∂y = zero(z)
```
devmotion, line 50: I think this should be
```suggestion
        ∂y = zero(w)
```
?

simsurace: Yes, will fix in next commit

simsurace:
I needed to modify the `frule` in order to make this pass. Since the derivative wrt. `x` is infinite, the finite difference tests would fail. I therefore replaced the `test_frule` and `test_rrule` by manual tests of just the `y` derivative.

simsurace:
I modified the wrong frule, apologies.

src/chainrules.jl
```diff
@@ -38,14 +44,24 @@ end
...
     return Ω, ∂x, ∂y
 end
 function ChainRulesCore.frule((_, Δx, Δy), ::typeof(xlog1py), x::Real, y::Real)
     Ω, ∂x, ∂y = _Ω_∂_xlog1py(x, y)
-    ΔΩ = muladd(∂x, Δx, ∂y * Δy)
+    if iszero(Δx)
+        ΔΩ = ∂y * Δy
+    else
+        ΔΩ = muladd(∂x, Δx, ∂y * Δy)
+    end
```
devmotion, line 64: Can you explain this change? Maybe you did but I'm still confused. Why do we want that `Δx` acts always like a strong zero? Shouldn't that only be the case when we pass `ZeroTangent` etc. or `false`?

simsurace: Well, otherwise the `-Inf` will result in `ΔΩ` being a `NaN`, which is different from what the `rrule` produces. Actually if the policy is to avoid NaNs, using `muladd` seems to make this harder in general. But I don't have a very good understanding of these matters as they relate to other considerations in the ecosystem. Maybe we can ask someone else for a second review.

devmotion: That's a general issue in ChainRules currently (there exist some issues as well IIRC). If you want to avoid `NaN`s usually you have to pass `ZeroTangent`/`NoTangent`/some other `AbstractTangent`/`false`, i.e., anything that is a strong zero. For instance:
```julia
julia> using ChainRules, ChainRulesCore

julia> frule((NoTangent(), 0.0), log, 0.0)
(-Inf, NaN)

julia> frule((NoTangent(), ZeroTangent()), log, 0.0)
(-Inf, ZeroTangent())

julia> frule((NoTangent(), NoTangent()), log, 0.0)
(-Inf, NoTangent())

julia> frule((NoTangent(), false), log, 0.0)
(-Inf, 0.0)

julia> frule((NoTangent(), 0.0), log1p, -1.0)
(-Inf, NaN)

julia> frule((NoTangent(), ZeroTangent()), log1p, -1.0)
(-Inf, ZeroTangent())

julia> frule((NoTangent(), NoTangent()), log1p, -1.0)
(-Inf, NoTangent())

julia> frule((NoTangent(), false), log1p, -1.0)
(-Inf, 0.0)
```

simsurace: I see, thanks for clearing that up. So I think it is better to revert this change and use `ZeroTangent()` in the test.

devmotion (approved):
OK, let's get this in! It's definitely an improvement and tests pass finally :slightly_smiling_face: We can always add more branches later if needed.

PR #62, opened 2023-01-25
Make ChainRulesCore, ChangesOfVariables, and InverseFunctions weak dependencies

devmotion:
By doing so, on Julia >= 1.9 support for these packages will be loaded only when the packages are loaded by the user (or some other package). Some more info about weak dependencies: https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)

The PR is based on https://github.com/JuliaMath/SpecialFunctions.jl/pull/421 and https://github.com/JuliaStats/StatsFuns.jl/pull/153.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LogExpFunctions"
 uuid = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 authors = ["StatsFun.jl contributors, Tamas K. Papp <tkpapp@gmail.com>"]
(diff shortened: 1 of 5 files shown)


devmotion:
It seems to work, e.g. https://github.com/JuliaStats/LogExpFunctions.jl/actions/runs/4010878874/jobs/6887860584#step:6:140

tpapp (approved):
LGTM, thanks for doing this.

PR #63, opened 2023-02-05
tweak how some packages are loaded in extensions

KristofferC:
Due to https://github.com/JuliaLang/julia/issues/48533, it might have to be necessary to disable packages loading other dependencies other than their weak dependencies and the parent package. The rest of the packages have to be transitively loaded from the parent package. This does this fix for the packages loaded in the extensions.

cc @vtjnash


devmotion (approved):
Oops, I did not know that it should be done in this way.

KristofferC:
This is a new change (or rather upcoming) so no one knew heh.

tpapp (approved):
Thanks!

PR #64, opened 2023-02-09
refactor: prefix extensions with package name

ven-k:
- this will ensure extension names are unique
- whenever there is another package with same ext name, building sysimage fails
- ex: ChangeOfVariables has its own ChainRulesCoreExt.jl

Ex: https://github.com/SciML/DiffEqBase.jl/tree/master/ext follows this for above mentioned reasons

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,9 +17,9 @@ ChangesOfVariables = "9e997f8a-9a97-42d5-a9f1-ce6bfc15e2c0"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 
 [extensions]
-ChainRulesCoreExt = "ChainRulesCore"
-ChangesOfVariablesExt = "ChangesOfVariables"
-InverseFunctionsExt = "InverseFunctions"
+LogExpFunctionsChainRulesCoreExt = "ChainRulesCore"
+LogExpFunctionsChangesOfVariablesExt = "ChangesOfVariables"
+LogExpFunctionsInverseFunctionsExt = "InverseFunctions"
 
 [compat]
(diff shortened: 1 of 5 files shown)


devmotion:
Is this an official requirement that is documented? Or just something SciML started to do? I think we would want to follow official guidelines and not change something preemptively if possibly it is changed/fixed in Pkg before Julia 1.9 is released.

ven-k:
- Not sure if it will be fixed by Julia 1.9.
- ATM, for building sysimages it is necessary. See this issue: https://github.com/JuliaLang/PackageCompiler.jl/issues/767

devmotion:
Copied from https://github.com/JuliaMath/ChangesOfVariables.jl/pull/13#issuecomment-1424844482:

> Personally, I would not make changes to the extensions too quickly and without approval from the Pkg developers since there does not even exist an official Julia version that supports extensions yet and the design, limitations, and restrictions are still changing (see e.g. https://github.com/JuliaLang/julia/issues/48533 and the PRs linked therein). Additionally, the official documentation for these extensions (https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) does not use prefixes but the current setup.

tpapp:
While it was great that a few packages went ahead to experiment with this feature, I think this can wait until things stabilize and we don't need to follow up temporary workarounds.

devmotion (approved):
Kristoffer thinks it is a reasonable change, regardless of potential upstream fixes. I trust him and therefore support the PR.

tpapp (approved):
@ven-k: Thanks!

@devmotion: I agree, will merge.

PR #65, opened 2023-02-20
Support IrrationalConstants 0.2

devmotion:
IrrationalConstants changed the type of the irrational constants to fix type piracy: https://github.com/JuliaMath/IrrationalConstants.jl/pull/19


tpapp (approved):
Thanks!

PR #70, opened 2023-06-01
add `logabssinh`

### LINKED ISSUE #69: `logabssinh`
LogExpFunctions already has `logcosh`. normally, when those terms appear, there is also `log(abs(sinh))` terms. In my specific field, that term appear in natural-gas properties:
image

logsinh can be expressed in terms of `LogExpFunctions.log1mexp`:
```
function logabssinh(x::Real)
    abs_x = abs(x)
    return abs_x + log1mexp(- 2 * abs_x) - IrrationalConstants.logtwo
end
```
Is it ok to have this function here?

longemen3000:
closes #69 . Implementation, chainrules and tests done

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LogExpFunctions"
 uuid = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
 authors = ["StatsFun.jl contributors, Tamas K. Papp <tkpapp@gmail.com>"]
-version = "0.3.23"
+version = "0.3.24"
 
 [deps]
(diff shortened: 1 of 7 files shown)


devmotion (commented):
Looks mainly good, thank you!

Can you add it to the docs and bump the version number?

devmotion (commented):
One additional point that I missed yesterday: Can you also add a test for the ChainRules definition in the respective test file?

devmotion (approved):
Looks good to me!

oscardssmith:
One last thing is that this will be pretty inaccurate when the output is near zero. I'm not sure if it matters, but I figured I would bring it up.

longemen3000:
in my particular case, the more interesting points are near x = 1, but it is a valid point.

PR #77, opened 2023-11-11
logstatsexp

cossio:
Adds the functions logmeanexp, logvarexp, logstdexp


src/logstatsexp.jl
```diff
@@ -0,0 +1,38 @@
+"""
+    logmeanexp(A; dims=:)
+
+Computes `log.(mean(exp.(A); dims))`, in a numerically stable way.
+"""
+function logmeanexp(A::AbstractArray; dims=:)
+    R = logsumexp(A; dims)
```
longemen3000, line 7: `logsumexp(A; dims)`, that syntax was not available in 1.0 I think, try using `dims=dims`

ParadaCarleton, line 7: Can you try giving everything more descriptive variable names? (Bing AI or GPT-4  are helpful for this IME.)

ParadaCarleton (changes requested):

src/logstatsexp.jl
```diff
@@ -0,0 +1,38 @@
...
+    N = length(A) ÷ length(R)
+    return R .- log(N)
```
devmotion, line 9: This will cause undesired promotions and allocations.

src/logstatsexp.jl
```diff
@@ -0,0 +1,38 @@
...
+end
+
+"""
+    logvarexp(A; dims=:)
+
+Computes `log.(var(exp.(A); dims))`, in a numerically stable way.
+"""
+function logvarexp(
+    A::AbstractArray; dims=:, corrected::Bool=true, logmean=logmeanexp(A; dims)
```
devmotion, line 18: We make sure to use a special single pass algorithm for logsumexp for performance reasons, so it seems a bit unfortunate to iterate over the data multiple times and precompute logmean by default. Could we use a single pass algorithm for the variance as well?

src/logstatsexp.jl
```diff
@@ -0,0 +1,38 @@
...
+)
+    R = logsumexp(2logsubexp.(A, logmean); dims)
+    N = length(A) ÷ length(R)
+	if corrected
+		return R .- log(N - 1)
```
devmotion, line 23: Again this causes undesired promotions and allocations.

test/logstatsexp.jl
```diff
@@ -0,0 +1,30 @@
...
+using Statistics: mean, var, std
+using LogExpFunctions: logmeanexp, logvarexp, logstdexp
+
+@testset "logmeanexp, logvarexp" begin
+    A = randn(5,3,2)
+    for dims in (2, (1,2), :)
+        @test logmeanexp(A; dims) ≈ log.(mean(exp.(A); dims))
+        for corrected in (true, false)
+            @test logvarexp(A; dims, corrected) ≈ log.(var(exp.(A); dims, corrected))
+            @test logstdexp(A; dims, corrected) ≈ log.(std(exp.(A); dims, corrected))
```
longemen3000: The `;dims` error is still here

src/logstatsexp.jl
devmotion: It seems you adressed my comment about peomotions but not the allocations.

src/logstatsexp.jl
devmotion: The unnecessary allocations are still present.

longemen3000:
I don't have an strong opinion on this, but Is it possible to implement this as an extension? This package is a really low level one, and adding a new dependency could have effects on downstream packages?

longemen3000:
`Statistics`, while a Stdlib, it is a new dependency

src/logstatsexp.jl
```diff
@@ -0,0 +1,38 @@
+"""
+    logmeanexp(A; dims=:)
```
devmotion, line 2: This docstring indicates that the functions works for arbitrary iterators (as `logsumexp`) whereas actually it is only defined for arrays. Can you fix the docstring or, preferably, add more general definitions?

longemen3000:
Ohhh, nvm, my error

ParadaCarleton:
I think we can merge now and work on optimization later. 

Honestly, it's just a huge pain trying to optimize for allocations without having Transducers.jl, laziness, or good functional primitives in Julia.

devmotion:
@ParadaCarleton as far as I know you've never contributed to this package, so I really think that you should let the maintainers of this package decide when a PR is ready and can be merge. In my opinion, being a member of JuliaStats doesn't mean that you are able and should merge PRs in repos that you haven't contributed to and/or are not familiar with. For instance, there are many JuliaStats packages where I don't think it's appropriate for me to merge PRs.

src/logstatsexp.jl
devmotion:
```julia
R .- log(N - 1)
```
creates a new array. But if `R` is e.g. of type `Array` this causes unnecessary allocations.

devmotion:
Clearly, we should not over-optimize this PR. But at the same time we should hold off merging PRs if reviewer comments that can be addressed without too much effort are not resolved. For instance, I think currently the following things are missing in this PR:
- Definitions for non-`AbstractArray` inputs (it seems the same algorithm could be applied?)
- Avoiding allocating `R .- log(N)` etc. when `R` is an `Array` and broadcasting when `R` is a scalar
- Tests with non-`Float64` inputs (these would have uncovered the promotion issues)

src/logstatsexp.jl
theogf: @devmotion you might argue that unless there's a way to modify `logsumexp` to directly take `log(N)` into account, an in-place operation might create issues with AD systems which do not support it?

tpapp:
I am closing this because of no activity. If there is interest, please bump here and I will reopen. That said, a fresh start would be preferred, addressing all the concerns here.

cossio:
@tpapp @devmotion Retrying at #120

Issue #79, opened 2024-02-14
xexpy(0, 1000) should be zero


cossio:
```julia
julia> xexpy(0, 1000)
NaN
```

I think `xexpy` can handle this better and return zero?

devmotion:
Why should it be 0? At least the current implementation agrees with the docstring.

tpapp:
I also think that we should make `xexpy` finite if `y` is finite and `x` is zero. The purpose of writing a function is precisely to handle these kind of cases.

devmotion:
Fixed by #80.

PR #80, opened 2024-02-14
fix `xexpy(0, 1000)`

oscardssmith:
now it equals 0.

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -62,7 +62,7 @@ end
 """
 $(SIGNATURES)
 
-Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf` or x == 0 and y is finite.
+Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf` or if `x == 0` and `y` is finite.
 
 ```jldoctest
 julia> xexpy(1.0, -Inf)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -62,7 +62,7 @@ end
 """
 $(SIGNATURES)
 
-Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf`.
+Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf` or x == 0 and y is finite.
 
 ```jldoctest
 julia> xexpy(1.0, -Inf)
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -72,7 +72,7 @@ julia> xexpy(1.0, -Inf)
 function xexpy(x::Real, y::Real)
     expy = exp(y)
     result = x * expy
-    return iszero(expy) && !isnan(x) ? zero(result) : result
+    return (iszero(x) && isfinite(y)) || (iszero(expy) && !isnan(x)) ? zero(result) : result
 end
 
 # The following bounds are precomputed versions of the following abstract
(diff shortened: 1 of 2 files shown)


oscardssmith:
tests added.

devmotion:
The PR does not match what `xexpy` is intended to return: https://github.com/JuliaStats/LogExpFunctions.jl/issues/79#issuecomment-1944223377

oscardssmith:
@devmotion I don't understand what you mean?

devmotion:
`xexpy(0, Inf)` is supposed to (as documented in the docstring) return `0 * exp(Inf)` which is `NaN`. To me it seems there is no bug to fix here.

To elaborate a bit more, `xexpy` and `xlogy` are supposed to satisfy the invariant `xexpy(a, b) = xlogy(exp(b), exp(a))`: If `b = -Inf`, or equivalently `exp(b) = 0`, then `xexpy(a, b) = xlogy(exp(b), exp(a)) = 0`; otherwise they return `a * exp(b) = exp(b) * log(exp(a))`.

oscardssmith:
This PR maintains `xexpy(0, Inf) == NaN`. The difference is in `xexpy(0, 10000)` where previously `exp(10000)` overflowed.

devmotion:
> Why? This requirement is not documented.

The behaviour of each function is documented which implicitly documents the invariance.

> I would also note that xlogy(1e100, 1) == 0, while xlogy(exp(1000), exp(0)) gives NaN only because the intermediate exp(1000) overflows.

My comment was supposed to be understood as a exact/mathematical description of the intended functions and their relation, disregarding floating-point issues such as overflow. That is, the invariance is for the mathematical functions underlying `xlogy` and `xexpy`,  but not their Julia implementations.

oscardssmith:
In that case, isn't this PR correct? This PR leaves xexpy(0, Inf) as NaN.

devmotion (commented):
I missed the check for `isfinite(y)`, so I think generally the PR is fine. It still disagrees with the docstring though since due to the change `xexpy(x, y)` is not always `x * exp(y)` if `y != -Inf` anymore. Can you adjust the docstring (e.g., by changing ``if `y == -Inf` `` to ``if `y == -Inf` or if `x == 0.0` and `y` is finite.``) and maybe also add the `xexpy(0, 1000)` example as a doctest?

oscardssmith:
I interpret the docstring as describing the function mathematically, in which case it is already correct.

devmotion:
Generally, the docstrings in LogExpFunctions mention explicitly if the computation deviates from the given Julia code ("preserving numerical accuracy", "carefully evaluated", "avoiding over/underflow" ...), so IMO the docstring should either mention the additional special case or contain a comment that indicates that the computation is not just `x * exp(y)`.

oscardssmith:
Ideally this code could avoid more overflow to deal with cases like `xexpy(floatmin(Float64), 1000.0)`, but that is rather difficult.

src/basicfuns.jl
```diff
@@ -62,7 +62,7 @@ end
 """
 $(SIGNATURES)
 
-Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf`.
+Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf` or x == 0 and y is finite.
```
devmotion, line 65: To improve the display in the docs:
```suggestion
Return `x * exp(y)` for `y > -Inf`, or zero if `y == -Inf` or if `x == 0` and `y` is finite.
```

devmotion (approved):

tpapp (approved):

Issue #36, opened 2022-03-06
clarify ChainRulesCore.jl integration


tpapp:
The README should make it clear that

1. all functions have working forward- and reverse rules, implemented via ChainRulesCore.jl

2. pull requests should implement and test these, too

PR #82, opened 2024-05-03
Add `loglogistic`, `logitexp`, `log1mlogistic` and `logit1mexp`

andrewjradcliffe:
This takes advantage of `LogExpFunctions`'s accurate implementations of `log1pexp` and `log1mexp`, combined with negation in the log-odds domain to provide more accurate and less expensive implementations of the function compositions.

As recommended [here](https://github.com/JuliaRegistries/General/pull/105931), I am submitting these.

Precedent for `loglogistic` and `log1mlogistic` is their existence in [Stan](https://mc-stan.org/docs/functions-reference/real-valued_basic_functions.html#composed-functions); it is natural to include their inverses.


src/basicfuns.jl
```diff
@@ -441,3 +441,72 @@ $(SIGNATURES)
...
+
+#=
+this uses the identity:
+
+log(logistic(x)) = -log(1 + exp(-x))
+=#
+"""
+$(SIGNATURES)
+
+The natural logarithm of the `logistic` function, computed more
```
devmotion, line 453: To be consistent with the style of the other docstrings and the Julia style guide, we should use the imperative form, e.g.
```suggestion
Compute the natural logarithm of the `logistic` function, computed more
```

Additionally, I think I'd prefer if the docstring would just state the expression that is computed, eg `log(logistic(x))` (with an additional comment that evaluation is performed more carefully and more performant). I don't think it's too helpful to try to describe the expression in words.

src/basicfuns.jl
```diff
@@ -441,3 +441,72 @@ $(SIGNATURES)
...
+carefully and with fewer calls than than the composition
+`log(logistic(x))`.
+
+Its inverse is the [`logitexp`](@ref) function.
+"""
+loglogistic(x::AbstractFloat) = -log1pexp(-x) #
```
devmotion, line 459: Why only `AbstractFloat`? I think we could just use
```suggestion
loglogistic(x::Real) = -log1pexp(-x)
```

src/basicfuns.jl
```diff
@@ -441,3 +441,72 @@ $(SIGNATURES)
...
+loglogistic(x::T) where {T<:Real} = -log1pexp(-convert(promote_type(Float64, T), x))
```
devmotion, line 460: I don't think this is needed - and it would cause undesired promotions of eg `Float32`.
```suggestion
```

src/basicfuns.jl
```diff
@@ -441,3 +441,72 @@ $(SIGNATURES)
...
+
+#=
+this uses the identity:
+
+logit(exp(x)) = log(exp(x) / (1 + exp(x))) = log(exp(x)) - log(1 - exp(x))
+=#
+"""
+$(SIGNATURES)
+
+The logit of the exponential of `x`, computed more carefully and
```
devmotion, line 470: Similarly, I' äd suggest using the imperative form and just stating the expression of what it returns.

src/basicfuns.jl
```diff
@@ -441,3 +441,72 @@ $(SIGNATURES)
...
+with fewer function calls than `logit(exp(x))`
+
+Its inverse is the [`loglogistic`](@ref) function.
+"""
+logitexp(x::Real) = x - log1mexp(x)
```
devmotion, line 475: Should this be done more carefully to avoid loss of precision when both terms are approximately equal?

src/basicfuns.jl
```diff
@@ -441,3 +441,72 @@ $(SIGNATURES)
...
+
+log(logistic(-x)) = -log(1 + exp(x))
+
+that is, negation in the log-odds domain.
+=#
+
+"""
+$(SIGNATURES)
+
+The natural logarithm of the 1 minus the inverse logit function,
```
devmotion, line 488: Same comments here 🙂

src/basicfuns.jl
```diff
@@ -441,3 +441,72 @@ $(SIGNATURES)
...
+
+this uses the same identity as `logitexp`, followed by negation on the
+log-odds scale, i.e. -logit(exp(x)) = log(1 - exp(x)) - log(exp(x))
+
+=#
+
+"""
+$(SIGNATURES)
+
+The logit of 1 minus the exponential of `x`, computed more carefully
```
devmotion, line 506: And here 🙂

src/basicfuns.jl
```diff
@@ -441,3 +441,69 @@ $(SIGNATURES)
...
+=#
+"""
+$(SIGNATURES)
+
+Return `log(logistic(x))`, computed more carefully and with fewer calls
+than the naive composition of functions.
+
+Its inverse is the [`logitexp`](@ref) function.
+"""
+loglogistic(x::AbstractFloat) = -log1pexp(-x)
```
devmotion, line 458:
```suggestion
loglogistic(x::Real) = -log1pexp(-x)
```
should be sufficient here (the line below should be removed, see https://github.com/JuliaStats/LogExpFunctions.jl/pull/82/files#r1589704384).

src/basicfuns.jl
andrewjradcliffe: Separate dispatch for `AbstractFloat` and all other `T<:Real` was a specific design decision, the logic for which is as follows:

Negation of `x` is necessary, but is only guaranteed to have the correct behavior for floating point types. In particular, unsigned integers will always produce nonsense results for any value other than `0`. The same applies to the typemin for all signed integers, though, that affects only 1 value, rather than 2^n - 1 values in the unsigned case. Rational numbers would replicate this problem, but `Base` throws an error when one attempts the pathological negations.

Notably, the undesired promotions only occur if the `AbstractFloat` dispatch is removed; as designed, it does no promotion of floating point types.

An alternate design is:
```julia
loglogistic(x::Real) = -log1pexp(-x)
loglogistic(x::T) where {T<:Integer} = loglogistic(convert(promote_type(Float64, T), x))
loglogistic(x::T) where {U<:Integer, T<:Rational{U}} = loglogistic(convert(promote_type(Float64, T), x))
```

I can see the motivation for this (`loglogistic(x::Real) = -log1pexp(-x)`) signature, so it looks like the alternate design should be used. Perhaps the explicit `convert` and `promote_type`s could be replaced with `float`, which, while potentially ambiguous, might be closer to what users expect.

devmotion: I guess generally most functions (in this package and other stats packages) assume that the additive inverse exists, so I assume using unsigned integers will cause problems all over the place. If you are worried about it here, I think you should just use `loglogistic(x::Real) = -log1pexp(-float(x))`. Promoting with an explicit `Float64` won't be correct in general (eg as mentioned above `Float16` or `Float32` must not be converted to `Float64`).

andrewjradcliffe: None of the definitions above promote `Float16` or `Float32` to `Float64`, only integers and rationals would be converted. See the example below

```julia
f(x::AbstractFloat) = x
f(x::T) where {T<:Real} = convert(promote_type(Float64, T), x)

@assert typeof(f(Float16(1.0))) === Float16
```

Moreover, `loglogistic(x::Real) = -log1pexp(-float(x))` is not necessarily correct for non-`Base` types which are `<:Real` but are not floating point types (e.g. [unums](https://en.wikipedia.org/wiki/Unum_(number_format)).

I'd rather err toward the least intrusive set of assumptions, while maintaining correctness, even if it leads to 2 more lines of code.

> most functions (in this package and other stats packages) assume that the additive inverse exists, so I assume using unsigned integers will cause problems all over the place.

Why not make the robust choice? 1 line of code is a very small burden to pay for elimination of a whole class of potential puzzles. Strictly speaking, a function should not accept unsigned integers if it just produces nonsense results.

src/basicfuns.jl
andrewjradcliffe: You're right -- this can be done more carefully using `logexpm1`.
However, as illustrated in the figures, `logexpm1` itself suffers a loss of precision around `-log(2)`.

Absolute error is ok.
logitexp_abserr

But units in the last place error is a problem around `-log(2)`
logitexp_ulperr

src/basicfuns.jl
andrewjradcliffe: Ok, so I thought I found a way to satisfy both the needs:
1) the AD abstractions, which really want a single function signature, which should be `f(x::Real)`
2) correctness wrt integer and rational arguments, without warts or quirks

The answer is straightforward: hide the necessary dispatch as a helper function, so that the public interface can satisfy the requirements AD backends expect.
```julia
loglogistic(x::Real) = _loglogistic(x)
_loglogistic(x::Real) = -log1pexp(-x)
_loglogistic(x::Union{Integer, Rational}) = _loglogistic(float(x))
```

Surely we can pay 2 extra lines of code to guarantee correctness wrt all integer and rational arguments, no?

src/basicfuns.jl
```diff
@@ -441,3 +441,74 @@ $(SIGNATURES)
...
+$(SIGNATURES)
+
+Return `log(logistic(x))`, computed more carefully and with fewer calls
+than the naive composition of functions.
+
+Its inverse is the [`logitexp`](@ref) function.
+"""
+loglogistic(x::Real) = _loglogistic(x)
+_loglogistic(x::Real) = -log1pexp(-x)
+_loglogistic(x::Union{Integer, Rational}) = _loglogistic(float(x))
```
devmotion, line 460: I'm still not sure whether unsigned integers are worth any consideration given that you will run into problems with them in many (most?) calculations and definitions in JuliaStats, JuliaMath, etc. typically do not handle them in a special way - but if we care then it would be easier to just define
```suggestion
loglogistic(x::Real) = -log1pexp(-float(x))
```
This would cover also cases such as e.g. unitful integer numbers that are not covered by `Union{Integer, Rational}` and for floating points such as `x::Float64` or `x::Float32` it would be equivalent to `-log1pexp(-x)`.

src/basicfuns.jl
```diff
@@ -441,3 +441,74 @@ $(SIGNATURES)
...
+$(SIGNATURES)
+
+Return `logit(exp(x))`, computed more carefully and with fewer calls than
+the naive composition of functions.
+
+Its inverse is the [`loglogistic`](@ref) function.
+"""
+logitexp(x::Real) = _logitexp(x)
+_logitexp(x::Real) = -logexpm1(-x)
+_logitexp(x::Union{Integer, Rational}) = _logitexp(float(x))
```
devmotion, line 477:
```suggestion
logitexp(x::Real) = -logexpm1(-float(x))
```

src/basicfuns.jl
```diff
@@ -441,3 +441,74 @@ $(SIGNATURES)
...
+$(SIGNATURES)
+
+Return `logit(1 - exp(x))`, computed more carefully and with fewer calls than
+the naive composition of functions.
+
+Its inverse is the [`log1mlogistic`](@ref) function.
+"""
+logit1mexp(x::Real) = _logit1mexp(x)
+_logit1mexp(x::Real) = logexpm1(-x)
+_logit1mexp(x::Union{Integer, Rational}) = _logit1mexp(float(x))
```
devmotion, line 514:
```suggestion
logit1mexp(x::Real) = logexpm1(-float(x))
```

ext/LogExpFunctionsChainRulesCoreExt.jl
```diff
@@ -187,4 +187,9 @@ end
 ChainRulesCore.@scalar_rule(cloglog(x), (-inv((1 - x) * log1p(-x)),))
 ChainRulesCore.@scalar_rule(cexpexp(x), (-xexpx(-exp(x)),))
 
+ChainRulesCore.@scalar_rule(loglogistic(x::Real), (logistic(-x),))
+ChainRulesCore.@scalar_rule(log1mlogistic(x::Real), (-logistic(x),))
+ChainRulesCore.@scalar_rule(logitexp(x::Real), (inv(1 - exp(x)),))
+ChainRulesCore.@scalar_rule(logit1mexp(x::Real), (-inv(1 - exp(x)),))
+
```
devmotion, line 194: IMO these are not needed - AD handles the functions already fine since rules for `log1pexp` etc. are defined.
```suggestion
```

andrewjradcliffe: As directed, these definitions have been removed.

src/basicfuns.jl
andrewjradcliffe: I opted for the `float` solution, as this is simplest and, furthermore, replicates the handling of integer and rational arguments in Julia `Base`.

> I'm still not sure whether unsigned integers are worth any consideration

SpecialFunctions.jl (another package I contributed to) converts most arguments via `float`, hence, it handles both the unsigned integer cases as well as the signed integer (i.e. `typemin(T<:Signed)`) edge case.

Use of `float` to convert integers and rationals to a floating point type for use in special functions is common throughout `Base`. Unless integers and rationals are converted, then a function interface defined as `x::Real` cannot be claimed. On the other hand, if the intention is to guarantee reasonable results only for floating point numbers, then this should clearly be stated in the package description. However, I am of the opinion that this package should follow the behavior of `Base` and `SpecialFunctions`, whereby integers and rationals are converted to floating point numbers internally.

Unless conversion is enforced, this leads to some obviously incorrect results, e.g., `logistic(typemin(Int)) == 1.0`, but, clearly, should be `0.0`.

src/basicfuns.jl
andrewjradcliffe: As this can only be resolved by improving the ulp error of `logexpm1` around `log(2)`, I see no reason for this to gate merging of this PR.

andrewjradcliffe:
As 
- all tests (included the added tests) are passing
- documentation has been updated
- the extension interfaces modified to support these 4 functions
- and tests of the package extensions have been added
please let me know of any additional changes which must be made for this PR to be merged.

ext/LogExpFunctionsChainRulesCoreExt.jl
andrewjradcliffe: Looking through the issues, [#36](https://github.com/JuliaStats/LogExpFunctions.jl/issues/36#issue-1160600914) seems to indicate that these functions should support `ChainRulesCore` forward and reverse rules.

src/basicfuns.jl
tpapp: I think we could check if the argument is in a neighborhood of -log(2), and then just use a first-order approximation. My quick calculations suggest that the local derivative is 2 but someone should check this.

tpapp (approved):
LGTM, as far as I can see all requests have been addressed. Thanks for the fine work!

andrewjradcliffe:
May we clarify: include or exclude `ChainRulesCore` forward/reverse rules?

Last item to adjust, then can merge.

andrewjradcliffe:
It would be best to not let this bit rot. Moreover, it is easiest to make changes when it is in recent memory.

@devmotion could you clarify with @tpapp whether to include or exclude ChainRulesCore forward/reverse rules?

Once clarified, I can either revert the last commit or leave it in place. Should I then assume that there would be no further obstacles to merging?

tpapp:
Feel free to remove them. Thanks for the ping!

andrewjradcliffe:
Reverted `ChainRulesCore` support. Should be ready for merge -- @devmotion and @tpapp

test/basicfuns.jl
```diff
@@ -468,3 +468,93 @@ end
     @test cexpexp(-Inf) == 0.0
     @test cexpexp(0) == (ℯ - 1) / ℯ
 end
+
+@testset "loglogistic: $T" for T in (Float16, Float32, Float64)
+    lim1 = T === Float16 ? -14.0 : -50.0
+    lim2 = T === Float16 ? -10.0 : -37.0
+    xs = T[Inf, -Inf, 0.0, lim1, lim2]
+    for x in xs
+        @test loglogistic(x) == log(logistic(x))
```
devmotion, line 477: If we test multiple types `T`, I think it would be good to also check that - as desired - the results are of type `T`.

andrewjradcliffe: Anything else, @devmotion?

devmotion (approved):

Issue #83, opened 2024-07-14
Add "a" parameter to softplus


DominiqueMakowski:
The "a" parameter can control the smooth function, which is useful in the context of regression modelling. 

- See https://link.springer.com/article/10.1007/s00362-023-01509-x

![image](https://github.com/user-attachments/assets/3ca805a3-3119-4f51-a637-f1b2df14984a)

![image](https://github.com/user-attachments/assets/bf138e92-7f0f-4f44-a859-748210b7d143)

Would it be possible to add it as an optional argument (defaulting to 1) ?

tpapp:
Fixed by #85.

PR #85, opened 2024-09-03
Add "a" parameter to softplus()  #83

DominiqueMakowski:
Following up the issues related to an exp link-function (https://github.com/TuringLang/Turing.jl/issues/2310), it reinforced the idea that a softplus link could actually be a good alternative. However, I feel like implementing its generalized version (#83) would be key (useful when modelling small parameters), so here my shot at it.

image


src/basicfuns.jl
```diff
@@ -165,9 +165,14 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
 This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref).
 
+The generalized `softplus` function (Wiemann et al., 2024) takes an additional optional parameter `a` that control
```
devmotion, line 168: I assume there exist earlier references for this function?

src/basicfuns.jl
```diff
@@ -165,9 +165,14 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
 This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref).
 
+The generalized `softplus` function (Wiemann et al., 2024) takes an additional optional parameter `a` that control 
+the approximation error with respect to the linear spline. It defaults to `a=1.0`, in which case the softplus is 
+equivalent to `log1pexp`.
+
```
devmotion, line 171: I think this should rather be a documentation of the generalized softplus function but not be part of `log1pexp` (possibly linked though).

src/basicfuns.jl
```diff
@@ -255,10 +260,22 @@ Return `log(exp(x) - 1)` or the “invsoftplus” function.  It is the inverse o
 [`log1pexp`](@ref) (aka “softplus”).
 """
 logexpm1(x::Real) = x <= 18.0 ? log(_expm1(x)) : x <= 33.3 ? x - exp(-x) : oftype(exp(-x), x)
-logexpm1(x::Float32) = x <= 9f0 ? log(expm1(x)) : x <= 16f0 ? x - exp(-x) : oftype(exp(-x), x)
+logexpm1(x::Float32) = x <= 9.0f0 ? log(expm1(x)) : x <= 16.0f0 ? x - exp(-x) : oftype(exp(-x), x)
```
devmotion, line 263: No need to change anything here?
```suggestion
logexpm1(x::Float32) = x <= 9f0 ? log(expm1(x)) : x <= 16f0 ? x - exp(-x) : oftype(exp(-x), x)
```

src/basicfuns.jl
```diff
@@ -255,10 +260,22 @@ Return `log(exp(x) - 1)` or the “invsoftplus” function.  It is the inverse o
...
+    end
+    return log1pexp(a * x) / a
+end
+
+function invsoftplus(y; a::Real=1.0)
+    if a == 1.0
+        return logexpm1(y)
+    end
+    return logexpm1(a * y) / a
+end
```
devmotion, line 277: These functions are type unstable (e.g. if `x isa Float32` and `a isa Float64`). I think it would be better to define (and document) something like
```suggestion
softplus(x::Real) = log1pexp(x)
softplus(x::Real, a::Real) = log1pexp(a * x) / a

invsoftplus(y::Real) = logexpm1(y)
invsoftplus(y::Real, a::Real) = logexpm1(a * y) / a
```

devmotion (commented):
I'm not sure how widely use this variant is (and whether there are other commonly used alternatives, the issue mentions also Liu and Ferber 2016?). If it's added, we should make to sure to test it and to also add support for it in the ChainRules, InverseFunctions, and ChangesOfVariables extensions.

src/basicfuns.jl
```diff
@@ -165,9 +165,12 @@ Return `log(1+exp(x))` evaluated carefully for largish `x`.
 This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
 transformation, being a smooth approximation to `max(0,x)`. Its inverse is [`logexpm1`](@ref).
 
+This is also called the ["softplus"](https://en.wikipedia.org/wiki/Rectifier_(neural_networks))
+transformation (in its default parametrization, see [`softplus`](@ref)), being a smooth approximation to `max(0,x)`. 
+
 See:
  * Martin Maechler (2012) [“Accurately Computing log(1 − exp(− |a|))”](http://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf)
-"""
+ """
```
devmotion, line 173:
```suggestion
"""
```

activate.jl
devmotion, line 1: Can you remove this file again, I guess it was committed unintentionally?

src/basicfuns.jl
```diff
@@ -257,8 +260,22 @@ Return `log(exp(x) - 1)` or the “invsoftplus” function.  It is the inverse o
...
+the approximation error with respect to the linear spline. It defaults to `a=1.0`, in which case the softplus is 
+equivalent to [`log1pexp`](@ref).
+
+See:
+ * Wiemann, P. F., Kneib, T., & Hambuckers, J. (2024). Using the softplus function to construct alternative link functions in generalized linear models and beyond. Statistical Papers, 65(5), 3155-3180.
+"""
+softplus(x::Real) = log1pexp(x)
+softplus(x::Real, a::Real) = log1pexp(a * x) / a
+
+invsoftplus(y::Real) = logexpm1(y)
```
devmotion, line 276: Maybe also add a docstring to this function?

devmotion (commented):
I think the remaining items here are:
- Include the new docstrings in the documentation
- Add tests for `softplus` and `invsoftplus`
- Add support for InverseFunctions for `softplus` and `invsoftplus` and test it
- Add support for ChangesOfVariables for `softplus` and `invsoftplus` and test it

I think ChainRules support should not be needed since `log1pexp` and `log1mexp` are already supported, and we can expect AD to differentiate through the remaining parts of the functions.

src/basicfuns.jl
DominiqueMakowski: I went through [Liu and Farber](https://[arxiv.org/pdf/1706.03609](https://arxiv.org/pdf/1706.03609)) to double-check

From my understanding (ML is not my  field), they validate "noisy softplus" as an improvement over other activation functions for neurons in NNs. 

image

However, it seems like they  named the `a` parameter sigma. Their plot looks similar but different in terms of values (?)

image

> I'm not sure how widely use this variant 

I share your concern here, I'm also careful not to add niche features to such a base package and add maintaining burden. 
I can't say how commonly the generalized version is already used, its development  seems fairly recent.
However, I can see its usefulness in quite a lot of cases: the default softplus only becomes close to identity after x > 2, and from experience we often do model parameters smaller than that (typical sigmas in my neuroscience/psychology are like between 0 and 1), so using adjust softplus links would make sense in these contexts. I suppose it's a tradeoff between the complexity of the feature and its (potential) usage

src/basicfuns.jl
DominiqueMakowski: A separate one or describe it in the above docstring?

devmotion: A separate one I think, but it would be useful to cross-reference them I assume.

DominiqueMakowski:
> Add support for InverseFunctions / ChangesOfVariables

Can you clarify?

DominiqueMakowski:
*Kind bump*

devmotion:
Sorry, I missed your previous comment.

> Add support for InverseFunctions / ChangesOfVariables

Since this PR adds new functions, we should also add definitions of `InverseFunctions.inverse` to https://github.com/JuliaStats/LogExpFunctions.jl/blob/289114f535827c612ce10c01b8dec9d3a55e4d15/ext/LogExpFunctionsInverseFunctionsExt.jl and definitions of `ChangesOfVariables.with_logabsdet_jacobian` to https://github.com/JuliaStats/LogExpFunctions.jl/blob/289114f535827c612ce10c01b8dec9d3a55e4d15/ext/LogExpFunctionsChangesOfVariablesExt.jl. Additionally, we could add definitions of `ChainRulesCore.frule` and `ChainRulesCore.rrule` - but in principle AD should "just work" since all other involved functions are known to ChainRules.

DominiqueMakowski:
I am not sure how to specify the ChangesofVariables one 🤔

DominiqueMakowski:
*Kind bump*

DominiqueMakowski:
Since there is no preexisting `ChangesOfVariables.with_logabsdet_jacobian` for softplus I'm really not sure what to write there

devmotion (commented):
I had re-reviewed the PR but apparently forgotten to submit the review on GH.

devmotion (approved):

tpapp (approved):
LGTM

PR #86, opened 2024-11-16
add `logabstanh`

oscardssmith:
as requested by @c05 on slack. This has 15 ULP for Float64, and 5 ULP for Float32.

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -149,6 +149,31 @@ end
 """
 $(SIGNATURES)
 
+Return `log(abs(tanh(x)))`, carefully evaluated without intermediate calculation of `tanh(x)`.
+
+The implementation ensures `logabstanh(-x) = logabstanh(x)`.
+"""
+function logabstanh(x::Real)
+    return log1p(-2/((exp(2abs(x))+1)))
+end
+function logabstanh(x::Float32)
+    abs_x = abs(x)
+    if abs_x < 0.0625f0
+        return log(abs_x) - x*x*(1f0/3)
+    end
+    return log1p(-2/((exp(2abs_x)+1)))
+end
+function logabstanh(x::Float64)
+    abs_x = abs(x)
+    if abs_x < 0x1p-5
+        return log(abs_x) + evalpoly(x*x, (0, -1/3, 7/90, -62/2835))
+    end
+    return log1p(-2/((exp(2abs_x)+1)))
+end
+
+"""
+$(SIGNATURES)
+
 Return `log(1+x^2)` evaluated carefully for `abs(x)` very small or very large.
 """
 log1psq(x::Real) = log1p(abs2(x))


src/basicfuns.jl
```diff
@@ -149,6 +149,31 @@ end
 """
 $(SIGNATURES)
 
+Return `log(abs(tanh(x)))`, carefully evaluated without intermediate calculation of `tanh(x)`.
+
+The implementation ensures `logabstanh(-x) = logabstanh(x)`.
+"""
+function logabstanh(x::Real)
+    return log1p(-2/((exp(2abs(x))+1)))
```
devmotion, line 157:
```suggestion
    return log1p(-2/(exp(2*abs(x))+1))
```

src/basicfuns.jl
```diff
@@ -149,6 +149,31 @@ end
...
+end
+function logabstanh(x::Float32)
+    abs_x = abs(x)
+    if abs_x < 0.0625f0
+        return log(abs_x) - x*x*(1f0/3)
```
devmotion, line 162: Any reason to not just use
```suggestion
        return log(abs_x) - x^2/3
```
?

src/basicfuns.jl
```diff
@@ -149,6 +149,31 @@ end
...
+    end
+    return log1p(-2/((exp(2abs_x)+1)))
```
devmotion, line 164:
```suggestion
    return log1p(-2/(exp(2*abs_x)+1))
```

src/basicfuns.jl
```diff
@@ -149,6 +149,31 @@ end
...
+end
+function logabstanh(x::Float64)
+    abs_x = abs(x)
+    if abs_x < 0x1p-5
+        return log(abs_x) + evalpoly(x*x, (0, -1/3, 7/90, -62/2835))
+    end
+    return log1p(-2/((exp(2abs_x)+1)))
```
devmotion, line 171:
```suggestion
    return log1p(-2/(exp(2*abs_x)+1))
```

devmotion (commented):
Thank you! Can you add tests and update the docs? Probably should also add a ChainRules definition.

src/basicfuns.jl
devmotion, line 157: Probably the safer default would be just

```suggestion
    return log(abs(tanh(x)))
```

devmotion, line 157: As indicated by the special cases close to 0 for Float32 and Float64, maybe a safer default would be
```suggestion
    return log(abs(tanh(x)))
```

oscardssmith: I think the version I wrote stays more accurate overall but I'll do a test

devmotion: I don't think it's possible to do such an analysis? You can't realistically test any subtype of `Real`.

oscardssmith: for any `AbstractFloat` `log(tanh(x))` will be inaccurate for large x since `tanh(x)` will be very close to 1.

devmotion:
> Can you add tests and update the docs? Probably should also add a ChainRules definition.

PR #87, opened 2024-12-10
Update CI setup, add dependabot, and update Julia compat to 1.10

devmotion:
Since it turned out that tests fail with Julia 1.0.0 (Julia version installed with `min`), I update the Julia compat to 1.10, the current LTS.


tpapp (approved):
thanks!

PR #94, opened 2025-07-05
improve `log1pmx`, add `Float32` implimentation

oscardssmith:
Through some clever use of Remez.jl (and some testing to better limit where the fallback is appropriate), we can remove almost all of the branches from the Float64 implementation and add a similar (but slightly faster) Float32 version.

This PR needs some CI tests added before merging, but I believe the algorithm is correct at this point.

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -352,6 +352,7 @@ function _log1pmx_ker(x::T) where T <: Union{Float32, Float64}
              0.1337701340211177,
              0.11201972567415432,
              0.143418239946679)
+    end
     w = evalpoly(t, p)
     hxsq = x*x/2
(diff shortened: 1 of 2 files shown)


oscardssmith:
@devmotion can you clarify what tests you want added? We already seem to have tests for Float32.

tpapp:
See #96 for an update that reproduces this. It is unrelated to your PR, but I think it would be great to fix it first, merge, and then get back to this.

tpapp:
@devmotion: can you please clarify what other tests would be needed? I think that tightening the bounds on the previous ones is fine.

devmotion:
I was mainly referring to the first comment:

> This PR needs some CI tests added before merging, but I believe the algorithm is correct at this point.

PR #96, opened 2025-07-09
make tests have their own environment

tpapp:
Introduces a separate test environment to reproduce the problem in #94 (which is unrelated to that PR), then adds a fix by rearranging the branches and testing for infinites early.

I still don't understand why the previous setup was not using the latest ForwardDiff.jl, but this one does.


oscardssmith:
Can you revert the test Project changes? IMO the `[extras]` way of managing versions works better (and `TestEnv.jl` can be used to activate the test environment), since the separate Project.toml approach is easy to have issues where someone updates one but not the other, leading to issues.

oscardssmith:
I don't especially care. I think this should merge.

devmotion:
I think the correct fix for mysterious test environments would be to add compat entries for test environments. This is what a currently missing Aqua test would require as well.

src/logsumexp.jl
```diff
@@ -114,21 +114,23 @@ _logsumexp_onepass_op((xmax, r)::Tuple{<:Number,<:Number}, x::Number) =
...
+    # ``r = ∑ᵢ exp(xᵢ - xmax)`` over the same elements.
     _xmax, _r = if isnan(x) || isnan(xmax)
         # ensure that `NaN` is propagated correctly for complex numbers
         z = oftype(x, NaN)
         z, r + exp(z)
     else
         real_x = real(x)
         real_xmax = real(xmax)
-        if real_x > real_xmax
+        if isinf(real_x) && isinf(real_xmax) && (real_x * real_xmax) > 0
```
devmotion, line 127: I think this change should be benchmarked. IIRC the performance of `logsumexp` was very sensitive to even minor changes of the implementation. To avoid the multiplication one might want to use
```suggestion
        if isinf(real_x) && isinf(real_xmax) && sign(real_x) == sign(real_xmax)
```

test/Project.toml
```diff
@@ -0,0 +1,13 @@
...
+ChainRulesTestUtils = "cdddcdb0-9152-4a09-a978-84456f9df70a"
+ChangesOfVariables = "9e997f8a-9a97-42d5-a9f1-ce6bfc15e2c0"
+FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
+JET = "c3a54625-cd67-489e-a8e7-0a5a0ff4e31b"
+LogExpFunctions = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
+OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
devmotion, line 13: IMO test dependencies should also be specified with compat entries and this file should be updated with CompatHelper.

PR #97, opened 2025-08-01
improve `_log1pmx_ker` internal documentation on the polynomial

nsajko:
Note which function the polynomial approximation approximates in a comment in the source code.

Might be useful if someone decided to search for a slightly different polynomial.

### DIFF
--- a/src/basicfuns.jl
+++ b/src/basicfuns.jl
@@ -337,6 +337,8 @@ end
 # parameters foudn via Remez.jl, specifically:
 # g(x) = evalpoly(x, big(2)./ntuple(i->2i+1, 50))
 # p = T.(Tuple(ratfn_minimax(g, (1e-3, (.425/(.425+2))^2), 8, 0)[1]))
+# A closed form for the Taylor series, after a variable change, is
+# 2 * (atanh(x) - x) / x^3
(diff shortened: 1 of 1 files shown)


tpapp (approved):
thanks!

src/basicfuns.jl
```diff
@@ -353,6 +353,11 @@ function _log1pmx_ker(x::T) where T <: Union{Float32, Float64}
              0.11201972567415432,
              0.143418239946679)
     end
+    # Mathematically equivalent to:
```
devmotion, line 356: IMO if one wants to be precise, it's not "mathematically equivalent" but rather an approximation.

devmotion (commented):
I don't really understand the purpose of the PR - the polynomial and how it was found (with Remez.jl) is already documented in https://github.com/JuliaStats/LogExpFunctions.jl/blob/30b15e31a4ad42e002948ae786a4a20442ff6e61/src/basicfuns.jl#L334-L339.

The comment explains that the polynomial is an approximation of $$\sum_{i=0}^{49} 2x^i/(2i + 3)$$ which is a truncation of the power series $$\sum_{i=0}^{\infty} 2x^i/(2i + 3) = 2 (atanh(x) - x) / x^3$$.

Issue #102, opened 2025-08-02
`log2mexp(x)` accuracy vanishes around the right edge of the domain, at `log(2)`


nsajko:
The error in ULPs (here for `log2mexp(x::Float64)`) has a spike at the right endpoint of the domain:

<img width="1502" height="480" alt="Image" src="https://github.com/user-attachments/assets/dbedad1a-09a5-492b-9f56-7978dcd708ff" />

The base of the spike:

<img width="1488" height="482" alt="Image" src="https://github.com/user-attachments/assets/d5d59c98-bcac-4fb0-83d9-341e340f231b" />

~~A problem is that the doc string for some reason promises the exact current implementation, so fixing this bug requires a breaking change.~~

tpapp:
@nsajko, if you are planning to make a fix, I can wait with 1.0 (no pressure).

PR #103, opened 2025-08-06
relax the doc string of `log2mexp`

nsajko:
Breaking: prevent the doc string from promising the exact current implementation, to enable possible accuracy improvements in the future.

Updates #102


tpapp (approved):
thanks!

PR #104, opened 2025-08-17
add `logabstanh` function

nsajko:
Revive PR #86, but without the `Float32`-specific implementation and without the `Float64`-specific implementation.

Closes #86


nsajko:
> without the `Float32`-specific implementation and without the `Float64`-specific implementation

My reasoning is that these are merely optimizations, so they can be considered separately in a future PR.

PR #106, opened 2025-08-22
add `logabstanh` fast paths for `Float32`, `Float64`

oscardssmith:


src/basicfuns.jl
```diff
@@ -231,6 +231,21 @@ function logabstanh(x::Real)
     end
 end
 
+function logabstanh(x::Float32)
+    a = abs(x)
+    if abs_x < 0.0625f0
```
nsajko, line 236:
```suggestion
    if a < 0.0625f0
```

oscardssmith:
getting this much faster than the default while maintaining 3 ULPs is pretty hard and not obviously worthwhile.

PR #114, opened 2026-05-29
Add source path for LogExpFunctions in docs/Project.toml

devmotion:
Possibly an alternative to #113


tpapp (approved):
I prefer this one over #113

PR #120, opened 2026-06-03
logstatsexp v2

### LINKED ISSUE #81: Add logmeanexp
Hi, 

I noticed that R and Python have `logmeanexp`. I think this would be a simple, yet useful addition.  I would be willing to make a PR.  The basic function would be 

```
function logmeanexp(X::AbstractArray{<:Number})
    return logsumexp(X) - log(length(X))
end
```

Some additional logic would be needed to handled the `dims` keyword.

cossio:
Adds the functions logmeanexp, logvarexp, logstdexp

A second attempt at #77 . The content is the same, but with some cleanup addressing comments at #77 (with agent help). @tpapp 

Fixes #81

### DIFF
--- a/ext/LogExpFunctionsChainRulesCoreExt.jl
+++ b/ext/LogExpFunctionsChainRulesCoreExt.jl
@@ -152,66 +152,54 @@ function ChainRulesCore.rrule(::typeof(logsumexp), x::AbstractArray{<:Real}; dim
 end
 
 function ChainRulesCore.frule((_, Δx), ::typeof(logmeanexp), x::AbstractArray{<:Real}; dims=:)
-    Ω = logmeanexp(x; dims=dims)
+    Ω = logmeanexp(x; dims)
     n = length(x) ÷ length(Ω)
-    ΔΩ = sum(exp.(x .- Ω) .* Δx; dims=dims) ./ n
+    ΔΩ = sum(exp.(x .- Ω) .* Δx; dims) ./ n
     return Ω, ΔΩ
 end
 function ChainRulesCore.rrule(::typeof(logmeanexp), x::AbstractArray{<:Real}; dims=:)
-    Ω = logmeanexp(x; dims=dims)
+    Ω = logmeanexp(x; dims)
     n = length(x) ÷ length(Ω)
-    ∂x = exp.(x .- Ω) ./ n
-    project_x = ChainRulesCore.ProjectTo(x)
-    function logmeanexp_pullback(Ω̄)
-        x̄ = ChainRulesCore.InplaceableThunk(
-            Δ -> Δ .+= Ω̄ .* ∂x,
-            ChainRulesCore.@thunk(project_x(Ω̄ .* ∂x)),
-        )
-        return ChainRulesCore.NoTangent(), x̄
-    end
-    return Ω, logmeanexp_pullback
+    return Ω, _∂x_pullback(exp.(x .- Ω) ./ n, x)
 end
 
 function ChainRulesCore.frule((_, Δx), ::typeof(logvarexp), x::AbstractArray{<:Real}; dims=:, corrected::Bool=true)
-    Ω = logvarexp(x; dims=dims, corrected=corrected)
-    ΔΩ = sum(_∂x_logvarexp(x, dims) .* Δx; dims=dims)
+    logmean = logmeanexp(x; dims)
+    Ω = logvarexp(x; dims, corrected, logmean)
+    ΔΩ = sum(_∂x_logvarexp(x, logmean, dims) .* Δx; dims)
     return Ω, ΔΩ
 end
 function ChainRulesCore.rrule(::typeof(logvarexp), x::AbstractArray{<:Real}; dims=:, corrected::Bool=true)
-    Ω = logvarexp(x; dims=dims, corrected=corrected)
-    ∂x = _∂x_logvarexp(x, dims)
-    project_x = ChainRulesCore.ProjectTo(x)
-    function logvarexp_pullback(Ω̄)
-        x̄ = ChainRulesCore.InplaceableThunk(
-            Δ -> Δ .+= Ω̄ .* ∂x,
-            ChainRulesCore.@thunk(project_x(Ω̄ .* ∂x)),
-        )
-        return ChainRulesCore.NoTangent(), x̄
-    end
-    return Ω, logvarexp_pullback
+    logmean = logmeanexp(x; dims)
+    Ω = logvarexp(x; dims, corrected, logmean)
+    return Ω, _∂x_pullback(_∂x_logvarexp(x, logmean, dims), x)
 end
 function ChainRulesCore.frule((_, Δx), ::typeof(logstdexp), x::AbstractArray{<:Real}; dims=:, corrected::Bool=true)
-    Ω = logstdexp(x; dims=dims, corrected=corrected)
-    ΔΩ = sum((_∂x_logvarexp(x, dims) ./ 2) .* Δx; dims=dims)
+    logmean = logmeanexp(x; dims)
+    Ω = logstdexp(x; dims, corrected, logmean)
+    ΔΩ = sum(_∂x_logvarexp(x, logmean, dims) ./ 2 .* Δx; dims)
     return Ω, ΔΩ
 end
 function ChainRulesCore.rrule(::typeof(logstdexp), x::AbstractArray{<:Real}; dims=:, corrected::Bool=true)
-    Ω = logstdexp(x; dims=dims, corrected=corrected)
-    ∂x = _∂x_logvarexp(x, dims) ./ 2
+    logmean = logmeanexp(x; dims)
+    Ω = logstdexp(x; dims, corrected, logmean)
+    return Ω, _∂x_pullback(_∂x_logvarexp(x, logmean, dims) / 2, x)
+end
+function _∂x_logvarexp(x::AbstractArray{<:Real}, logmean, dims)
+    d = x .- logmean
+    e = expm1.(d)
+    return (2 .* exp.(d) .* e) ./ sum(abs2, e; dims)
+end
+function _∂x_pullback(∂x, x)
     project_x = ChainRulesCore.ProjectTo(x)
-    function logstdexp_pullback(Ω̄)
+    function pullback(Ω̄)
         x̄ = ChainRulesCore.InplaceableThunk(
             Δ -> Δ .+= Ω̄ .* ∂x,
             ChainRulesCore.@thunk(project_x(Ω̄ .* ∂x)),
         )
         return ChainRulesCore.NoTangent(), x̄
     end
-    return Ω, logstdexp_pullback
-end
-function _∂x_logvarexp(x::AbstractArray{<:Real}, dims)
-    d = x .- logmeanexp(x; dims=dims)
-    e = expm1.(d)
-    return (2 .* exp.(d) .* e) ./ sum(abs2, e; dims=dims)
+    return pullback
 end
 
 # no rules for mutating functions currently:
--- a/ext/LogExpFunctionsChainRulesCoreExt.jl
+++ b/ext/LogExpFunctionsChainRulesCoreExt.jl
@@ -151,6 +151,69 @@ function ChainRulesCore.rrule(::typeof(logsumexp), x::AbstractArray{<:Real}; dim
     return Ω, logsumexp_pullback
 end
 
+function ChainRulesCore.frule((_, Δx), ::typeof(logmeanexp), x::AbstractArray{<:Real}; dims=:)
+    Ω = logmeanexp(x; dims=dims)
+    n = length(x) ÷ length(Ω)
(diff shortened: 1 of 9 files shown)


tpapp (changes requested):
Thanks @cossio for coming back to this.

I find the copilot-generated code a bit convoluted. Specifically,

1. I am not sure why it makes sense to accumulate the length *in general*, when array-like objects can compute it cheaply; the only context where that makes sense is a generic iterator with `Base.SizeUnknown()`.
2. I am not sure of the purpose of `_subtract_log_count`, shouldn't `-Inf - log(0)` just take care of the empty case?

Generally, the way I think could be coded is by defining an
1. initial state,
2. an elementwise op,
3. a finalizer that gives the results
for each of these.

Then an internal function could implement something like
```julia
function _reduce(op, a::AbstractArray{T}, dims)
    _finalize(reduce(_reduction(op), a; dims, init = _init(op, float(T)), a)
end
```
where `_finalize(::typeof(logmeanexp))` etc are defined. A trait mechanism could special-case the `Base.SizeUnknown()` iterator. 

Note that this is just a suggestion, you should code it the way you prefer. I just want to keep the implementation simple and easy to maintain.

Incidentally, `mean_and_std` and `mean_and_var` would make sense.

cossio:
Added `logmeanexp_and_logvarexp` and `logmeanexp_and_logstdexp`, which share a common pass and thus are roughly 1.5x faster than individual calls to `logmeanexp` and `logvarexp`.

src/logstatsexp.jl
```diff
@@ -0,0 +1,195 @@
...
+#
+# The mean is `logsumexp(X) - log(n)`. The variance uses the *centered* formula
+#
+#     log var = logsumexp(2 * logsubexp(xᵢ, logmean)) - log(n - corrected)
+#
+# i.e. the log of the sum of squared deviations `∑ᵢ (exp(xᵢ) - mean)²`, divided by the
+# count. Centering is essential for numerical stability: the raw-moment alternative
+# `logsubexp(∑exp(2xᵢ), (∑exp(xᵢ))²/n)` cancels catastrophically when the variance is
+# small relative to the mean (and can even overflow to `Inf` for nearly-equal inputs).
+# Computing the variance therefore needs the mean first, hence the two helpers below.
```
tpapp: Can it be done in a single pass though? Updating the mean online too.

cossio: Yes — it can be done in a single pass with an online (Welford-style) mean, but it turns out to be substantially slower, so I've kept the two-pass version.

I prototyped the online variant (track `logmean` and `logS = log ∑ᵢ (exp(xᵢ) − mean)²`, updating both per element) and benchmarked it against the current two-pass implementation. Both agree with `Statistics.var(exp.(x))` to rounding. `Float64`, BenchmarkTools, Julia 1.12.6:

```
       n     two-pass     Welford   ratio
      10      0.29 µs     0.78 µs    2.7x
     100      2.01 µs     7.94 µs    4.0x
    1000     19.3  µs    82.4  µs    4.3x
   10000    213    µs   838    µs    3.9x
  100000      2.26 ms     8.44 ms    3.7x
 1000000     22.6  ms    84.6  ms    3.7x
```

The single-pass version is ≈3.7–4.3× slower across the whole range — including arrays well out of cache, where its halved memory traffic should help it most. Two reasons:

1. **More transcendental work per element.** The centered two-pass does ≈3 `exp`/`log`-domain ops per element (the mean pass, plus one `logsubexp` and the running `logsumexp` in the variance pass). The online update needs ≈5: two `logsubexp` for the old/new deviations, a `logaddexp`/`logsubexp` for the mean update, a `logaddexp` for the running `S`, plus a `log(k)`.
2. **The online recurrence is serial.** Each step depends on the previous running mean, so it can't vectorize. The mean pass of the two-pass version is just `logsumexp`, which does vectorize.

So memory bandwidth never gets to dominate — the extra `exp`/`log` evaluations do. Given that, I've left the two-pass centered implementation in place. Happy to revisit if you'd prefer the single-pass form for other reasons.

tpapp: Thanks for the explanation and the benchmarks.

cossio: Disclaimer: the previous comment was written by Claude. I told Claude to do the benchmarks, and then intended to reply myself, but it just went ahead and also replied here on my behalf. Anyway I think the answer is clear (probably better than what I would've written).

cossio:
@tpapp friendly bump, anything else needed here ?

src/logstatsexp.jl
devmotion: This comment is just AI slop, isn't it? Everyone knows that the naive variance calculation should never be used, there's nothing special about the `log`/`exp` setting here.

src/logstatsexp.jl
```diff
@@ -0,0 +1,195 @@
...
+
+Compute `log(mean(exp, X))` in a numerically stable way.
+
+`X` should be an iterator of numbers.
+"""
+function logmeanexp(X)
+    # single pass, counting as we go: never traverses `X` twice, so single-use iterators
+    # (and ones that only report a length) are handled correctly.
+    lse, n = _logsumexp_count(identity, X)
+    return lse - _log_count(lse, n)
```
devmotion: Why using a separate function here? Just promote `n` and `lse` (already in the line above if not done internally in `_logsumexp_count`) and return `lse - log(n)`?

src/logstatsexp.jl
```diff
@@ -0,0 +1,195 @@
...
+end
+
+"""
```
devmotion, line 40: IMO it's questionable whether this simple function needs two docstrings.

src/logstatsexp.jl
```diff
@@ -0,0 +1,195 @@
...
+$(SIGNATURES)
+
+Compute `log.(mean(exp.(X); dims=dims))` in a numerically stable way.
+"""
+function logmeanexp(X::AbstractArray{<:Number}; dims=:)
+    dims isa Colon && isempty(X) && _throw_empty()
```
devmotion: This should just return `NaN`, not error.

src/logstatsexp.jl
```diff
@@ -0,0 +1,195 @@
...
+    lse = logsumexp(X; dims=dims)
+    return lse .- _log_count(lse, _reduced_count(X, lse))
```
devmotion: Also this here seems overly complicated? And in many cases causes unnecessary allocations...

src/logstatsexp.jl
```diff
@@ -0,0 +1,195 @@
...
+end
+
+"""
+$(SIGNATURES)
+
+Compute `log.(mean(exp.(X); dims=dims))` in a numerically stable way.
+"""
+function logmeanexp(X::AbstractArray{<:Number}; dims=:)
+    dims isa Colon && isempty(X) && _throw_empty()
+    lse = logsumexp(X; dims=dims)
```
devmotion:
```suggestion
    lse = logsumexp(X; dims)
```

src/logstatsexp.jl
```diff
@@ -0,0 +1,195 @@
...
+    # single pass, counting as we go: never traverses `X` twice, so single-use iterators
+    # (and ones that only report a length) are handled correctly.
+    lse, n = _logsumexp_count(identity, X)
+    return lse - _log_count(lse, n)
+end
+
+"""
+$(SIGNATURES)
+
+Compute `log.(mean(exp.(X); dims=dims))` in a numerically stable way.
```
devmotion:
```suggestion
Compute `log.(mean(exp.(X); dims))` in a numerically stable way.
```

src/logstatsexp.jl
```diff
@@ -0,0 +1,195 @@
...
+function logmeanexp(X::AbstractArray{<:Number}; dims=:)
+    dims isa Colon && isempty(X) && _throw_empty()
+    lse = logsumexp(X; dims=dims)
+    return lse .- _log_count(lse, _reduced_count(X, lse))
+end
+
+"""
+$(SIGNATURES)
+
+Compute `(log(mean(exp, X)), log(var(exp, X; corrected=corrected)))` in a numerically
```
devmotion:
```suggestion
Compute `(log(mean(exp, X)), log(var(exp, X; corrected)))` in a numerically
```

src/logstatsexp.jl
```diff
@@ -0,0 +1,195 @@
...
+stable way. Computing the two together is cheaper than calling [`logmeanexp`](@ref) and
+[`logvarexp`](@ref) separately, since the mean is reused to center the variance.
+
+`X` should be an iterator of real numbers.
+"""
+function logmeanexp_and_logvarexp(X; corrected::Bool=true)
+    xs = _materialize(X)
```
devmotion: Why? `logmeanexp` above accepts any iterator.

devmotion (commented):
IMO this is a terribly verbose and partly overly complicated PR - definitely subpar to any existing implementations in this package. I think this should be cleaned up quite a bit before being merged.

src/logstatsexp.jl
cossio: Done

src/logstatsexp.jl
cossio: Agree

src/logstatsexp.jl
cossio: That was because an iterator can be consumed on the first pass, and this function needs two passes.
I decided to remove these `logmeanexp_and_logvarexp` functions, to save review bandwidth. A future PR can focus on these functions if needed.

cossio:
@devmotion cleaned-up a bit more. Hopefully it's better now.

Also added AD rules.

src/logstatsexp.jl
```diff
@@ -0,0 +1,128 @@
...
+end
+_LogSqDev(x::AbstractArray, mean::AbstractArray) =
+    _LogSqDev{float(eltype(x)),ndims(x),typeof(x),typeof(mean)}(x, mean)
+Base.size(s::_LogSqDev) = size(s.x)
+Base.axes(s::_LogSqDev) = axes(s.x)
+Base.IndexStyle(::Type{<:_LogSqDev}) = IndexCartesian()
+Base.@propagate_inbounds function Base.getindex(s::_LogSqDev{T,N}, I::Vararg{Int,N}) where {T,N}
+    j = map((i, ax) -> ifelse(length(ax) == 1, first(ax), i), I, axes(s.mean))
+    return convert(T, _logsqdev_term(s.x[I...], s.mean[j...]))
+end
```
devmotion: Is this really necessary? Since Julia doesn't have any contracts for (arguably loosely defined) interfaces, it's just too easy to miss definitions or define things in inefficient ways. I've seen this happen in particular with custom array definitions quite a bit...

I somehow have a slightly hopeless feeling regarding this PR - it seems to overcomplicate things and contain code that I (maybe incorrectly) assume was introduced by an AI system without controlling or thinking through the design. That gives me bad vibes when reviewing and will be annoying and unnecessarily complex for future maintainers.

ext/LogExpFunctionsChainRulesCoreExt.jl
```diff
@@ -151,6 +151,69 @@ function ChainRulesCore.rrule(::typeof(logsumexp), x::AbstractArray{<:Real}; dim
...
+function ChainRulesCore.frule((_, Δx), ::typeof(logmeanexp), x::AbstractArray{<:Real}; dims=:)
+    Ω = logmeanexp(x; dims=dims)
+    n = length(x) ÷ length(Ω)
+    ΔΩ = sum(exp.(x .- Ω) .* Δx; dims=dims) ./ n
+    return Ω, ΔΩ
+end
+function ChainRulesCore.rrule(::typeof(logmeanexp), x::AbstractArray{<:Real}; dims=:)
+    Ω = logmeanexp(x; dims=dims)
+    n = length(x) ÷ length(Ω)
+    ∂x = exp.(x .- Ω) ./ n
```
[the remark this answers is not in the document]
cossio: Both claims are false in Julia:
  - length(2.0) == 1 for a scalar Number — no MethodError.
  - sum([1,2,3]; dims=:) == 6 — Colon dims is supported and returns a scalar.

src/logstatsexp.jl
cossio: @devmotion This _LogSqDev construct was there to try to reduce allocations in logvarexp!. This was suggested by Claude, but I myself see no alternative way to reduce allocations `logvarexp!`.

If we are willing to accept a temporary allocation in `logvarexp!` then I can remove `_LogSqDev` it and I agree the code will be cleaner.

cossio: If you agree, I'll also drop support for iterators in `logvarexp`. The complexity is because it needs two passes over the iterator, so one needs to materialize the iterator anyways. I think with these two modifications will simplify the code substantially.

We then support iterators only for `logmeanexp` which can be done in one-pass. As far as I can see this is the cleanest solution.

cossio: I implemented the changes above:
- removed `_LogSqDev`
- dropped support for iterators in `logvarexp`

Could you take another look ? I think it's much cleaner now. Thanks

